PR #34, opened 2014-12-29
Fully support symlinks and submodules

sqs:
Lstat, Stat, and ReadDir now properly handle symlinks and (git) submodules. In addition:
- the Sys() method of the os.FileInfo for a symlink returns a SymlinkInfo that contains the symlink's destination
- the Sys() method of the os.FileInfo for a git submodule returns a SubmoduleInfo that contains the submodule's URL and commit ID

### DIFF
--- a/vcs/file_info.go
+++ b/vcs/file_info.go
@@ -1,0 +1,23 @@
+package vcs
+
+// ModeSubmodule is an os.FileMode mask indicating that the file is a
+// VCS submodule (e.g., a git submodule).
+const ModeSubmodule = 0160000
+
+// SubmoduleInfo holds information about a VCS submodule and is
+// returned in the FileInfo's Sys field by Stat/Lstat/ReadDir calls.
+type SubmoduleInfo struct {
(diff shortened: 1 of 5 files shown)


vcs/file_info.go
```diff
@@ -0,0 +1,23 @@
+package vcs
+
+// ModeSubmodule is an os.FileMode mask indicating that the file is a
+// VCS submodule (e.g., a git submodule).
+const ModeSubmodule = 0160000
```
shurcooL: @sqs, what's the reason that this submask is 3 bits rather than a single bit?

I'm seeing a lot of the code here does the following to check if something is a submodule:

```Go
if fi.Mode()&vcs.ModeSubmodule != 0 {
```

But when I tried that, I've run into false positives, which I fixed by doing:

```Go
if fi.Mode()&vcs.ModeSubmodule == vcs.ModeSubmodule {
```

Just checking, do you remember what the context here was?

sqs: Sorry, I don't remember.

PR #38, opened 2015-01-18
gitcmd: Fix ResolveRevision failing to return vcs.ErrRevisionNotFound on valid-looking revisions.

### LINKED ISSUE #37: gitcmd: ResolveRevision failing to return vcs.ErrRevisionNotFound on valid-looking revisions.
In trying to track down and resolve https://github.com/shurcooL/gtdo/issues/12, I think I've found the root cause in this repo.

The problem appears to be that `go-vcs/vcs/gitcmd` implementation of `ResolveRevision` fails to return `vcs.ErrRevisionNotFound` when the commit "looks" like a valid commit hash.

This situation often comes up when trying to do a `ResolveRevision` on a recently-pushed new commit that is not present in the `vcsstore` store because the repo needs to be updated. But `ResolveRevision` returns no error in that case, and my code doesn't try to update the repo, and when it tries to actually access that commit by getting a file system, reading from there obviously fails.

Thankfully, the tests are passing now (yay \o/) so that will make fixing this a lot easier. I will begin by making a PR that adds failing test cases, then we can discuss and agree on a good fix.

The current `gitcmd` implementation of `ResolveRevision` appears to simply do `git rev-parse {{.Rev}}`, but I _think_ it should be doing something like `git rev-parse --verify '{{.Rev}}^{commit}'` or so, but anyway, I'll look more into that after starting the PR.

shurcooL:
Work in progress PR to close #37, see there for context.

Currently only added failing testcases.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -106,7 +106,7 @@ func (r *Repository) ResolveRevision(spec string) (vcs.CommitID, error) {
 		return "", err
 	}
 
-	cmd := exec.Command("git", "rev-parse", spec)
+	cmd := exec.Command("git", "rev-parse", spec+"^{commit}")
 	cmd.Dir = r.Dir
 	stdout, stderr, err := dividedOutput(cmd)
 	if err != nil {
--- a/vcs/hg/repo.go
+++ b/vcs/hg/repo.go
@@ -80,7 +80,7 @@ func (r *Repository) ResolveRevision(spec string) (vcs.CommitID, error) {
 
 	rec, err := r.parseRevisionSpec(spec).Lookup(r.cl)
 	if err != nil {
-		if err == hex.ErrLength {
+		if err == hg_revlog.ErrRevNotFound || err == hex.ErrLength {
 			return "", vcs.ErrRevisionNotFound
 		}
 		return "", err
--- a/vcs/repository_test.go
+++ b/vcs/repository_test.go
@@ -205,24 +205,47 @@ func TestRepository_ResolveRevision_error(t *testing.T) {
 		spec    string
 		wantErr error
 	}{
-		"git libgit2": {
+		"git libgit2 testcase1": {
 			repo:    makeGitRepositoryLibGit2(t, gitCommands...),
 			spec:    "doesntexist",
 			wantErr: vcs.ErrRevisionNotFound,
 		},
-		"git cmd": {
+		"git cmd testcase1": {
 			repo:    makeGitRepositoryCmd(t, gitCommands...),
 			spec:    "doesntexist",
 			wantErr: vcs.ErrRevisionNotFound,
 		},
-		"hg": {
+		"hg testcase1": {
 			repo:    makeHgRepositoryNative(t, hgCommands...),
 			spec:    "doesntexist",
 			wantErr: vcs.ErrRevisionNotFound,
 		},
-		"hg cmd": {
+		"hg cmd testcase1": {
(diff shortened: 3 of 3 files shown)


shurcooL:
There are currently 2 failures:

```
repository_test.go:256: git cmd testcase2: ResolveRevision: got <nil>, want revision not found
repository_test.go:256: hg testcase2: ResolveRevision: got revision not found, want revision not found
```

"hg testcase2" is also failing. I've investigated why, and the reason is that the error is coming from `github.com/beyang/hgo/revlog` package, where it's defined as:

```Go
var ErrRevNotFound = errors.New("revision not found")
```

So, it looks similar, but it's not `== vcs.ErrRevisionNotFound`.

I see two fixes, either change `hgo/revlog` to return vcs:

```Go
var ErrRevNotFound = vcs.ErrRevisionNotFound
```

But that would require for `hgo/revlog` to import `sourcegraph.com/sourcegraph/go-vcs/vcs`.

Alternatively, in `sourcegraph.com/sourcegraph/go-vcs/vcs/hg`, add the following check:

```diff
func (r *Repository) ResolveRevision(spec string) (vcs.CommitID, error) {
	if id, err := r.ResolveBranch(spec); err == nil {
		return id, nil
	}
	if id, err := r.ResolveTag(spec); err == nil {
		return id, nil
	}

	rec, err := r.parseRevisionSpec(spec).Lookup(r.cl)
	if err != nil {
-		if err == hex.ErrLength {
+		if err == hex.ErrLength || err == revlog.ErrRevNotFound {
			return "", vcs.ErrRevisionNotFound
		}
		return "", err
	}
	return vcs.CommitID(hex.EncodeToString(rec.Id())), nil
}
```

shurcooL:
Ok, fixed the "hg testcase2" failing testcase in 7f1581a, so just "git cmd testcase2" remaining.

From [git docs on rev-parse](http://git-scm.com/docs/git-rev-parse):

> If you want to make sure that the output actually names an object in your object database and/or can be used as a specific type of object you require, you can add "\^{type}" peeling operator to the parameter. For example, git rev-parse "$VAR^{commit}" will make sure $VAR names an existing object that is a commit-ish (i.e. a commit, or an annotated tag that points at a commit). To make sure that $VAR names an existing object of any type, git rev-parse "$VAR^{object}" can be used.

So I think the current `ResolveRevision` code should be changed from:

```Go
cmd := exec.Command("git", "rev-parse", spec)
```

To:

```Go
cmd := exec.Command("git", "rev-parse", spec+"^commit")
```

And that will fix the issue. Trying that now.

shurcooL:
Okay, that works to fix the added failing test case, but it's too aggressive and causes other previously passing tests to fail:

```
--- FAIL: TestRepository_BlameFile (0.94s)
	blame_test.go:98: git cmd: ResolveRevision("master") on base: revision not found
...
--- FAIL: TestRepository_ResolveBranch (1.29s)
	repository_test.go:74: git cmd: ResolveBranch: branch not found
```

Currently, `ResolveBranch` and `ResolveTag` of `gitcmd` do nothing more than call `ResolveRevision`, so that may need to change a little.

shurcooL:
Okay, I think `spec+"^{commit}"` may actually work. I am still getting a few failed tests locally, but that seems to be independent of the change. I get the same failed tests with `master` code.

So I'll try committing the `spec+"^{commit}"` change and see what Travis says.

shurcooL:
PTAL.

I think this is good to merge, pending review.

sqs:
LGTM. Thanks!

PR #39, opened 2015-02-17
Add "resolving dependencies" section to readme

MarkMcCaskey:


sqs:
Thanks!

The included Makefile has an `install` target that already does some of this (specifically, the `cd` and `make install` for git2go, but not the cloning). I think it would be best to add the additional cloning steps to the Makefile and then mention the Makefile targets that must be run in the README. That way, we don't have the same commands living in multiple places (which could cause issues if we forget to update one of the places, for example).

MarkMcCaskey:
Thanks for the feedback!  I really appreciate it!  I've updated the Makefile and README locally to reflect the changes you suggested, but I'm unsure of whether I should have it take care of installing libgit2 as well.
Thanks in advance!  I look forward to doing more work with Sourcegraph in the future!

Makefile
```diff
@@ -1,10 +1,16 @@
...
 install:
-	cd `go list -f '{{.Dir}}' github.com/libgit2/git2go` && make install
+	cd `go list -f '{{.Dir}}' github.com/libgit2/git2go` && git submodule update --init && make install && go install
 
 test:
 	go test -ldflags "-extldflags=-L"`go list -f '{{.Dir}}' github.com/libgit2/git2go`/vendor/libgit2/build ./...
 
 docker-test:
 	docker build -t go-vcs . && docker run go-vcs
+dep:
```
sqs: I think you can remove the `dep` target and modify the README to say "Run `go get ./...` to install dependencies." That way, we don't need to update the `dep` target when we add or remove dependencies.

MarkMcCaskey:
Alright, I think I finally have a working version.  Thanks for the help!  I'm new to Go, so the dependencies had me confused for a bit, but with help from you and Samer I think I have a pretty good grasp of it now.
As far as I can tell, it should be working.  Let me know if you notice anything wrong with it though.

Thanks,
Mark

sqs:
Awesome, thank you! :+1:

Feel free to make other contributions to our Go libs if you want to practice and get feedback while you're learning Go.

PR #44, opened 2015-04-17
Add BranchesOptions for Repository.Branches method, support getting behind/ahead counts.

shurcooL:
This is a breaking API change for `Branches` method of `Repository` interface. As any breaking API change, it is not done lightly, but needs to be done in order to support additional features for `Branches` method call. By using an options struct, future enhancements and pagination support will not incur additional API breakage.

- It changes the signature of `Branches` to be more like `Commits`, which accepts an `Options` struct where parameters may be specified that describe details of the `Branches` query.

- ~~It also adds a `total uint` return value, which will always be the total number of branches, even if fewer are returned (due to pagination, for example). Note that there is no pagination support at this time, but since it's likely to be added in the future, it's hopefully better to avoid a second API breakage at that time.~~

- Implement support for `BehindAheadBranch` in `vcs/gitcmd` implementation. If set to a non-empty string, it uses that branch name as the base branch, and gets relative behind/ahead commit counts for each branch returned.

### TODO.

- [ ] Add BehindAheadBranch support to `vcs/git` implementation.
- [x] Add tests.

### DIFF
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -320,7 +320,7 @@ func main() {
 			log.Fatalln("no supported vcs found in cwd")
 		}
 
-		repo, err := vcs.Open(r.Type().VcsType(), ".")
+		repo, err := vcs.Open(r.Type().VcsType(), r.RootPath())
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -329,12 +329,12 @@ func main() {
 		if len(args) == 1 {
 			opt.BehindAheadBranch = args[0]
 		}
-		branches, total, err := repo.Branches(opt)
+		branches, err := repo.Branches(opt)
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		fmt.Printf("# Branches (%d total):\n", total)
+		fmt.Printf("# Branches (%d total):\n", len(branches))
 		for _, b := range branches {
 			switch {
 			case b.Counts == nil:
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -14,9 +14,10 @@ import (
 	"strings"
 
 	"github.com/kr/text"
+	vcs2 "github.com/shurcooL/go/vcs"
 	"sourcegraph.com/sourcegraph/go-diff/diff"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs"
-	_ "sourcegraph.com/sourcegraph/go-vcs/vcs/git"
+	//_ "sourcegraph.com/sourcegraph/go-vcs/vcs/git"
 	_ "sourcegraph.com/sourcegraph/go-vcs/vcs/gitcmd"
 	_ "sourcegraph.com/sourcegraph/go-vcs/vcs/hg"
 )
@@ -309,23 +310,38 @@ func main() {
 		}
 
 	case "branches":
-		if len(args) != 0 {
-			log.Fatal("branches takes no arguments.")
+		if len(args) > 1 {
+			log.Fatal("branches takes 0 or 1 arguments.")
 		}
 
-		repo, err := vcs.Open("hg", ".")
+		// Open using go/vcs to figure out VCS type (git, hg).
+		r := vcs2.New(".")
+		if r == nil {
+			log.Fatalln("no supported vcs found in cwd")
+		}
+
+		repo, err := vcs.Open(r.Type().VcsType(), ".")
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		branches, err := repo.Branches()
+		var opt vcs.BranchesOptions
+		if len(args) == 1 {
+			opt.BehindAheadBranch = args[0]
+		}
+		branches, total, err := repo.Branches(opt)
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		fmt.Printf("# Branches (%d total):\n", len(branches))
+		fmt.Printf("# Branches (%d total):\n", total)
 		for _, b := range branches {
-			fmt.Printf("%s %s\n", b.Head, b.Name)
+			switch {
+			case b.Counts == nil:
+				fmt.Printf("%s %s\n", b.Head, b.Name)
+			case b.Counts != nil:
+				fmt.Printf("-%v | +%v | %s\n", b.Counts.Behind, b.Counts.Ahead, b.Name)
+			}
 		}
 	}
 }
--- a/vcs/git/repo.go
+++ b/vcs/git/repo.go
@@ -104,13 +104,13 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return "", vcs.ErrTagNotFound
 }
 
-func (r *Repository) Branches(_ vcs.BranchesOptions) (branches []*vcs.Branch, total uint, err error) {
+func (r *Repository) Branches(_ vcs.BranchesOptions) ([]*vcs.Branch, error) {
 	r.editLock.RLock()
 	defer r.editLock.RUnlock()
 
 	refs, err := r.u.NewReferenceIterator()
 	if err != nil {
-		return nil, 0, err
+		return nil, err
 	}
 
 	var bs []*vcs.Branch
@@ -120,7 +120,7 @@ func (r *Repository) Branches(_ vcs.BranchesOptions) (branches []*vcs.Branch, to
 			break
 		}
 		if err != nil {
-			return nil, 0, err
+			return nil, err
 		}
 		if ref.IsBranch() {
 			bs = append(bs, &vcs.Branch{Name: ref.Shorthand(), Head: vcs.CommitID(ref.Target().String())})
@@ -128,7 +128,7 @@ func (r *Repository) Branches(_ vcs.BranchesOptions) (branches []*vcs.Branch, to
 	}
 
 	sort.Sort(vcs.Branches(bs))
-	return bs, uint(len(bs)), nil
+	return bs, nil
 }
 
 func (r *Repository) Tags() ([]*vcs.Tag, error) {
--- a/vcs/git/repo.go
+++ b/vcs/git/repo.go
@@ -104,13 +104,13 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return "", vcs.ErrTagNotFound
 }
 
-func (r *Repository) Branches() ([]*vcs.Branch, error) {
+func (r *Repository) Branches(_ vcs.BranchesOptions) (branches []*vcs.Branch, total uint, err error) {
 	r.editLock.RLock()
 	defer r.editLock.RUnlock()
 
 	refs, err := r.u.NewReferenceIterator()
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
 	var bs []*vcs.Branch
@@ -120,7 +120,7 @@ func (r *Repository) Branches() ([]*vcs.Branch, error) {
 			break
 		}
 		if err != nil {
-			return nil, err
+			return nil, 0, err
 		}
 		if ref.IsBranch() {
 			bs = append(bs, &vcs.Branch{Name: ref.Shorthand(), Head: vcs.CommitID(ref.Target().String())})
@@ -128,7 +128,7 @@ func (r *Repository) Branches() ([]*vcs.Branch, error) {
 	}
 
 	sort.Sort(vcs.Branches(bs))
-	return bs, nil
+	return bs, uint(len(bs)), nil
 }
 
 func (r *Repository) Tags() ([]*vcs.Tag, error) {
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -150,23 +150,55 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return commitID, nil
 }
 
-func (r *Repository) Branches() ([]*vcs.Branch, error) {
+// branchCounts returns the behind/ahead commit counts information for branch, against base branch.
+func (r *Repository) branchCounts(branch, base string) (behind, ahead int, err error) {
+	cmd := exec.Command("git", "rev-list", "--count", "--left-right", base+"..."+branch)
+	cmd.Dir = r.Dir
+	out, err := cmd.Output()
+	if err != nil {
+		return 0, 0, err
+	}
+	behindAhead := strings.Split(strings.TrimSuffix(string(out), "\n"), "\t")
+	behind, err = strconv.Atoi(behindAhead[0])
+	if err != nil {
+		return 0, 0, err
+	}
+	ahead, err = strconv.Atoi(behindAhead[1])
+	if err != nil {
+		return 0, 0, err
+	}
+	return behind, ahead, nil
+}
+
+func (r *Repository) Branches(opt vcs.BranchesOptions) (branches []*vcs.Branch, total uint, err error) {
 	r.editLock.RLock()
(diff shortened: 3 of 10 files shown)


vcs/repository.go
```diff
@@ -140,10 +141,24 @@ type Diff struct {
 	Raw string // the raw diff output
 }
 
+// BranchesOptions specifies options for the list of branches returned by
+// (Repository).Branches.
+type BranchesOptions struct {
+	BehindAheadBranch string `json:",omitempty" url:",omitempty"` // Or "" to disable fetching behind/ahead counts.
```
shurcooL: Possible alternative names are `BaseBranch`, `BehindAheadBaseBranch`. I should also elaborate on what it does exactly in the docs.

cmd/go-vcs/go-vcs.go
```diff
@@ -309,23 +310,38 @@ func main() {
...
 
 	case "branches":
-		if len(args) != 0 {
-			log.Fatal("branches takes no arguments.")
+		if len(args) > 1 {
+			log.Fatal("branches takes 0 or 1 arguments.")
+		}
+
+		// Open using go/vcs to figure out VCS type (git, hg).
+		r := vcs2.New(".")
```
sqs: Instead of adding the whole shurcooL/go repo to this repo's dependencies, can you just extract this function? It would also be handy to have it work for all subcommands of the go-vcs command.

vcs/repository.go
```diff
@@ -140,10 +141,28 @@ type Diff struct {
...
 	Name string
 	Head CommitID
+
+	Counts *BehindAhead `json:",omitempty"` // Counts optionally contains the commit counts relative to specified branch.
+}
+
+// BehindAhead is a set of behind/ahead counts.
+type BehindAhead struct {
+	Behind int
+	Ahead  int
```
shurcooL: Maybe these should be made into `uint`, to avoid confusion. Their values are always non-negative. For example:

```
Behind: 4
Ahead: 17
```

That means target branch is 4 commits behind base branch, and 17 commits ahead.

cmd/go-vcs/go-vcs.go
```diff
@@ -309,23 +310,38 @@ func main() {
...
+		if r == nil {
+			log.Fatalln("no supported vcs found in cwd")
 		}
 
-		repo, err := vcs.Open("hg", ".")
+		repo, err := vcs.Open(r.Type().VcsType(), ".")
```
sqs: You don't need to make this change in the upstream lib, but I thought I'd mention it anyway. VcsType should be VCSType, according to [CodeReviewComments#Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

cmd/go-vcs/go-vcs.go
shurcooL: I just added it for testing, I didn't mean to keep it. I needed it because the repo I was testing with was "git", not "hg".

As far as extracting it into here, I can't extract just that function, since you'd need the entire `github.com/shurcooL/go/vcs` library. It's a relatively small library, but it's more than just one func. See for yourself:

https://sourcegraph.com/github.com/shurcooL/go@master/.tree/vcs

However, what I can (and want) to do is extract the entire `go/vcs` Go package into a standalone git repository. Would that be better?

Eventually, I'd like to merge the functionality of `shurcooL/go/vcs` together with `sourcegraph/go-vcs`, or rather make them more compatible and play better together. Right now it's a little awkward because they share some responsibilities but differ in other aspects.

vcs/repository.go
```diff
@@ -27,8 +27,9 @@ type Repository interface {
 	// ErrBranchNotFound if no such branch exists.
 	ResolveBranch(name string) (CommitID, error)
 
-	// Branches returns a list of all branches in the repository.
-	Branches() ([]*Branch, error)
+	// Branches returns a list of all branches in the repository, as well as
+	// the total number of branches.
+	Branches(BranchesOptions) (branches []*Branch, total uint, err error)
```
sqs: This may not be an issue for branches, but I found that having a `total` return value was a bad idea for commits. That's because it requires all implementations to perform an exact count, which is an expensive operation. I.e., if you are listing the most recent 10 commits, that's quick--but the API contract requires the implementation to iterate through ALL commits to perform an exact count (which can take 2-5 seconds on big repos).

Notice that GitHub doesn't return an exact count, I assume for this very reason. See https://github.com/golang/go/commits/master?page=2 at the bottom, where the page links are just "Older" and "Newer" and not a full page count.

Branches are obviously much fewer in number than commits but the same principle applies.

Possible solutions?

* Mention in the docs that the implementation is not required to perform an exact count if `total >= 50` and that callers should treat any value equal to or above 50 as a lower bound.
* Add an extra return value `exactTotal bool` that is false if the count is a lower bound. (Or make a single return value of `type count struct { n uint; exact bool}`.)
* Just return a `more bool` indicating whether there are any branches beyond those returned.
* Punt for now -- assume the number of branches will be small enough.

cmd/go-vcs/go-vcs.go
shurcooL: I am aware of my deviation from the Go style in this instance, and it is a valid point. Thanks!

Most of my personal Go code breaks that Go style suggestion (but only that one). I had some reasons for doing it. I may want to revision my decision about it in the future.

vcs/gitcmd/repo.go
```diff
@@ -150,23 +150,55 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
 			Name: strings.TrimPrefix(ref[1], "refs/heads/"),
 			Head: vcs.CommitID(ref[0]),
 		}
 	}
-	return branches, nil
+
+	// Fetch behind/ahead counts for each branch.
+	if opt.BehindAheadBranch != "" {
+		for i, branch := range branches {
+			behind, ahead, err := r.branchCounts(branch.Name, opt.BehindAheadBranch)
```
sqs: Consider parallelizing these operations. Even if invoking `git` is quick (~10msec on my desktop on local SSD is probably the best case), performing this operation for 10-20 branches would take 100-200msec serially.

vcs/gitcmd/repo.go
```diff
@@ -150,23 +150,55 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return commitID, nil
 }
 
-func (r *Repository) Branches() ([]*vcs.Branch, error) {
+// branchCounts returns the behind/ahead commit counts information for branch, against base branch.
+func (r *Repository) branchCounts(branch, base string) (behind, ahead int, err error) {
+	cmd := exec.Command("git", "rev-list", "--count", "--left-right", base+"..."+branch)
```
sqs: Use `checkSpecArgSafety` on `base` (and probably `branch` too) to make sure it doesn't begin with `--` (which is a possible attack vector).

vcs/repository.go
shurcooL: I see. That's a great point.

I think it makes sense to make `total` return value always present if and only if it's very inexpensive to return in all cases.

I can imagine that is not the case for commits. Unless the total number of commits is cached somewhere, computing it would require traversing the entire list of commits from branch tip to the root commit (one that has no more parents), one commit at a time. That's gotta be quite expensive, so it's wasteful to make the API have it as a mandatory return.

In that case, let's just remove it for now (for branches, it is currently redundant and unused information). Once we decide to add pagination support to branches, we can revisit and come up with an optimal solution. Does that sound okay?

vcs/git/repo.go
```diff
@@ -104,13 +104,13 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return "", vcs.ErrTagNotFound
 }
 
-func (r *Repository) Branches() ([]*vcs.Branch, error) {
+func (r *Repository) Branches(_ vcs.BranchesOptions) (branches []*vcs.Branch, total uint, err error) {
```
sqs: You can omit the `_`. It doesn't matter, but I didn't realize that was possible until a few months ago myself and thought I'd mention it in case you didn't know.

cmd/go-vcs/go-vcs.go
sqs: Yeah, extracting the package into its own repo would be great. Not a big deal either way, since this is in the command, not the lib (and Godeps users of this lib wouldn't pull in the cmd deps).

vcs/repository.go
sqs: Sounds good.

vcs/gitcmd/repo.go
shurcooL: Would it make sense to focus on performance and optimization in the `vcs/git` implementation, rather than `vcs/gitcmd`, which I would imagine is not used when performance is a concern?

Parallelizing this would improve response time if the system is not heavily loaded. But it may reduce performance if load is already near max, and it would complicate the implementation.

I'm working on the `vcs/git` implementation now and I look into parallelizing it there.

I'm not against it, just want to balance these things carefully.

sqs:
Looks great to me. Minor comments. Only important one is the `checkSpecArgSafety` note and perhaps the parallelization comment (which is optional--you can ignore that for now). Feel free to merge after you addressed.

vcs/git/repo.go
shurcooL: Thanks. I am aware it's possible to avoid it, but I purposefully included the blank identifier as an indicator that the parameter is completely unused.

vcs/gitcmd/repo.go
shurcooL: Thanks! This is important, I will do this.

vcs/repository.go
shurcooL: @sqs Thoughts on this?

sqs: Yeah, that would make it clearer actually. I notice that the Go stdlib uses signed ints in many places where the values are constrained to be nonnegative--perhaps to avoid type conversions. But I am all for additional clarity.

vcs/gitcmd/repo.go
sqs: Yeah, this is not high-priority. Sounds good.

vcs/repository.go
shurcooL: Right. I also think the added clarity is worth the possible additional type conversion.

Especially in this context where it may be less clear if positive/negative numbers will be used.

Will change!

vcs/gitcmd/repo.go
shurcooL: From my (can't guarantee to be thorough) testing, it's possible to create weird branch names, but not ones that start with a dash.

However, you may have a branch named `dev` and another branch literally named `refs/heads/dev`. If you use the short branch name in the comparison, `git` prints a warning:

```
$ git rev-list --count --left-right refs/heads/master...refs/heads/dev
warning: refname 'refs/heads/dev' is ambiguous.
0	1
```

But if you prefix the branch name with "refs/heads/", it's no longer ambiguous:

```
$ git rev-list --count --left-right refs/heads/master...refs/heads/refs/heads/dev
27	5
```

But then you are allowed to make a branch literally named "refs/heads/refs/heads/dev", which makes the above ambiguous again.

sqs: I don't think we should rely on the input arguments being safe for 2 reasons:

1. Someone might call this method from somewhere else that gets the arguments from user input, and not think about this issue.
2. `git show-ref --heads` might print out a warning message that is in part determined by user input (or the name of a file that an attacker could commit), and we might parse that warning message (it's probably going to stderr, but we might combine both, or it might go to stderr) as a branch name. I can't think of a way right now, but I can't guarantee it wouldn't happen. (E.g., suppose git warned if you had really long branch names that were deprecated and going to be forbidden in git 17.0 in the future. Say the branch name is `aaaaaaaaaaaaaaaaaaaasdf`. The warning message might be `warning: branch name is too long, truncating 'aaaasdf'`. Then if the branch was named like `aaaaaaaaaaaaaaa--foo`, we might parse that warning message as a branch named `--foo`. Totally hypothetical.)

And if it's impossible to create branch names that begin with a dash, then checking the args will never break the functionality on a valid branch name, so there's no downside.

shurcooL: I agree.

To clarify, when I made the observation that it didn't seem to be possible to create branch names beginning with a dash, I did not mean to imply we could assume that's always the case. After all, maybe it's possible to create branches that begin with a dash via alternative means.

However, I thought there may be a way to make the branch name safe (even if it begins with a dash, etc.) by prepending "refs/heads/" to it.

But that doesn't seem to guarantee anything, so I will do it alongside with the `checkSpecArgSafety` check.

sqs: Oh, got it. :+1:

shurcooL: Done in bf75e7a.

vcs/repository.go
shurcooL: Done in f220901.

I estimate a small chance (less than 10%) of needing to revert to `int` for some unexpected reason (this is git, after all) over the course of the next few years. Hopefully the extra clarity and type safety (unsigned ints for _count_ values) is worth that risk, but we'll see!

cmd/go-vcs/go-vcs.go
shurcooL: > Yeah, extracting the package into its own repo would be great.

I've created shurcooL/go#18 and I will look into addressing it eventually.

> Not a big deal either way, since this is in the command, not the lib (and Godeps users of this lib wouldn't pull in the cmd deps).

Agreed. The `cmd/go-vcs` is mainly used for testing/development, and the [installation instructions](https://github.com/sourcegraph/go-vcs#installing) specify:

```
go get sourcegraph.com/sourcegraph/go-vcs/vcs
```

Which means the dependencies of `go get sourcegraph.com/sourcegraph/go-vcs/cmd/go-vcs` will not be fetched, so it being a bigger repository will not affect those users.

That's why I'll leave it as is for now. It can be improved when doing next set of improvements/feature development for `go-vcs/vsc`.

vcs/repository.go
shurcooL: Done in dba9f86. Removed `total` return from the `Branches` method API.

shurcooL:
Addressed all comments, PTAL.

sqs:
LGTM!

shurcooL:
Great!

I will look at the remaining 2 TODOs tomorrow and merge afterwards. Since this is a breaking API change, some [dependent packages](https://godoc.org/sourcegraph.com/sourcegraph/go-vcs/vcs?importers) will need to be updated, but it'll be just a matter of doing:

```
gofmt -r 'Branches() -> Branches(vcs.BranchesOptions{}'
```

shurcooL:
Added test in 178fbedeaa825651bdf38f8372bc68d94a443c90.

shurcooL:
Okay, the `vcs/git` implementation is not done because it is a larger undertaking. We can defer doing it until it is needed.

I think this is good to proceed. The `vcs.BranchesOptions` struct makes it possible to extend options later without breaking API changes.

cmd/go-vcs/go-vcs.go
```diff
@@ -14,9 +14,10 @@ import (
 	"strings"
 
 	"github.com/kr/text"
+	vcs2 "github.com/shurcooL/go/vcs"
 	"sourcegraph.com/sourcegraph/go-diff/diff"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs"
-	_ "sourcegraph.com/sourcegraph/go-vcs/vcs/git"
+	//_ "sourcegraph.com/sourcegraph/go-vcs/vcs/git"
```
sqs: Can you add a TODO about this? Or even better, a GitHub issue or list in the readme of "differences between gitcmd and git impls".

sqs:
After that minor comment about adding a TODO, lgtm! :ship:

shurcooL:
Merging. :shipit:

As a reminder, this is a minor breaking API change (and these are needed occasionally to move forward), but updating is easy:

```
gofmt -r 'a.Branches() -> a.Branches(vcs.BranchesOptions{})'
```

PR #47, opened 2015-04-19
Improve git installation process.

### LINKED ISSUE #43: vcs/git: Installation process and/or instructions can be improved.
I've installed it recently, and found issues/problems along the way. I'll want to improve the process and/or instructions.

shurcooL:
- Mention that both `libssh2` and `libgcrypt` are hard prerequisites for
libgit2 implementation in advance of its installation.
- The Makefile "install" target was outdated (missing the `git checkout
next` step). In addition to that, it repeats the exact same steps as
described in Resolving Dependencies section, which is assumed to be
done by the time user gets to Installing section. I removed it since
it's unneeded and may cause confusion.
- Closes #43.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,10 @@ matrix:
     - go: tip
 
 before_install:
-  - mkdir -p $HOME/gopath/src/sourcegraph.com/sourcegraph
-  - mv $TRAVIS_BUILD_DIR $HOME/gopath/src/sourcegraph.com/sourcegraph/go-vcs
-  - export TRAVIS_BUILD_DIR=$HOME/gopath/src/sourcegraph.com/sourcegraph/go-vcs
+  - export GOPATH="$HOME/gopath" # Fix until https://github.com/travis-ci/travis-ci/issues/3439 is resolved.
+  - mkdir -p $GOPATH/src/sourcegraph.com/sourcegraph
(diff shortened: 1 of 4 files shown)


shurcooL:
Previously, the first commit made a larger change, see e07a34fe769a1355ec941ff6ff0fbfe87e0ca935.

However, I found on another OS X machine, manually specifying `#cgo CFLAGS: -I/usr/local/include` and `#cgo LDFLAGS: -L/usr/local/lib` was not needed for it to work, which is more in line with that I've expected. I will avoid adding those folders for now until I can have more conclusive evidence that it is needed.

shurcooL:
This is ready for review and merging, PTAL.

shurcooL:
> However, I found on another OS X machine, manually specifying `#cgo CFLAGS: -I/usr/local/include` and `#cgo LDFLAGS: -L/usr/local/lib` was not needed for it to work, which is more in line with that I've expected. I will avoid adding those folders for now until I can have more conclusive evidence that it is needed.

Yeah, I was right, it was a local OS X configuration issue. `/usr/local/{include,lib}` paths were not included in the list of searched paths, but normally they should be. Doing `xcode-select --install` on this Mac fixed it, as suggested at http://stackoverflow.com/questions/23905661/on-mac-g-fails-to-search-usr-local-include-and-usr-local-lib-by-default/26404710#26404710.

PR #48, opened 2015-04-21
define some types in protobuf (for use in go-sourcegraph, etc.)

sqs:


vcs/gen.go
```diff
@@ -0,0 +1,3 @@
+package vcs
+
+//go:generate protoc -I../../../.. -I ../../../../github.com/gogo/protobuf/protobuf -I. --gogo_out=. vcs.proto
```
shurcooL: I'm guessing there shouldn't be a space in the second `-I` parameter:

```
-I../../../..
-I ../../../../github.com/gogo/protobuf/protobuf
-I.
```

It may still work, but it's better to fix it for consistency.

PR #49, opened 2015-05-06
Add optional commit details to vcs.Branch.

shurcooL:
When calling vcs.Branches method, it is now possible to set IncludeCommit option to true. Doing so will include full commit details for each returned branch (i.e., the head commit for each branch).

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -196,6 +196,16 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 		}
 	}
 
+	if opt.IncludeCommit {
+		for i, branch := range branches {
+			c, err := r.getCommit(branch.Head)
+			if err != nil {
+				return nil, err
+			}
+			branches[i].Commit = c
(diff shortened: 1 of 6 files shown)


vcs/gitcmd/repo.go
```diff
@@ -277,10 +287,7 @@ func exitStatus(err error) int {
 	return 0
 }
 
-func (r *Repository) GetCommit(id vcs.CommitID) (*vcs.Commit, error) {
-	r.editLock.RLock()
-	defer r.editLock.RUnlock()
-
+func (r *Repository) getCommit(id vcs.CommitID) (*vcs.Commit, error) {
```
sqs: Add a doc comment to `getCommit` saying that `r.editLock` MUST be held by the caller.

sqs:
LGTM after that 1 comment about `getCommit` - PGAM (please go ahead and merge :smile:) afterwards

vcs/gitcmd/repo.go
shurcooL: Good idea, will do, thanks!

PR #50, opened 2015-05-06
Add support for file/dir history.

shurcooL:
- This is done by allowing to select only commits that modify a given path.
- The direction of this commit is largely based on research and findings from commit 2a71007acc73d25867c6e9e6a0e706b346b7336c. See it and its commit message for rationale.
- Minor fixes to ./cmd/go-vcs and improvements to code style consistency.

This is ready for initial review. /cc @sqs I may want to do some minor fixups and add tests tomorrow.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -343,12 +343,20 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
 		args = append(args, "--skip="+strconv.FormatUint(uint64(opt.Skip), 10))
 	}
 
+	if opt.Path != "" {
+		args = append(args, "--follow")
+	}
+
 	// Range
 	rng := string(opt.Head)
 	if opt.Base != "" {
 		rng += "..." + string(opt.Base)
 	}
 	args = append(args, rng)
+
+	if opt.Path != "" {
+		args = append(args, "--", opt.Path)
+	}
 
 	cmd := exec.Command("git", args...)
 	cmd.Dir = r.Dir
@@ -401,6 +409,12 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
 
 	// Count commits.
 	cmd = exec.Command("git", "rev-list", "--count", rng)
+	if opt.Path != "" {
+		// TODO: This doesn't include --follow flag because rev-list doesn't support it, so the number may be slightly off.
+		//       Consider removing the "total" return value from the API altogether, that's why I'm not trying to do more to produce an accurate number here.
+		//       Getting the total number of commits is an expensive operation and likely isn't needed, or deserves a separate dedicated API if needed.
+		cmd = exec.Command("git", "rev-list", "--count", rng, "--", opt.Path)
+	}
 	cmd.Dir = r.Dir
 	out, err = cmd.CombinedOutput()
 	if err != nil {
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -334,6 +334,9 @@ func isInvalidRevisionRangeError(output, obj string) bool {
 	return strings.HasPrefix(output, "fatal: Invalid revision range "+obj)
 }
 
+// commitLog returns a list of commits.
+//
+// The caller is responsible for doing checkSpecArgSafety on opt.Head and opt.Base.
 func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error) {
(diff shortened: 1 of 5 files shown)


vcs/gitcmd/repo.go
```diff
@@ -401,6 +409,12 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
 
 	// Count commits.
 	cmd = exec.Command("git", "rev-list", "--count", rng)
+	if opt.Path != "" {
+		// TODO: This doesn't include --follow flag because rev-list doesn't support it, so the number may be slightly off.
+		//       Consider removing the "total" return value from the API altogether, that's why I'm not trying to do more to produce an accurate number here.
+		//       Getting the total number of commits is an expensive operation and likely isn't needed, or deserves a separate dedicated API if needed.
```
shurcooL: I propose we make the breaking API change sooner, in order to improve the API (IMO), and remove `total` from `Commits`.

It belongs in a separate API call IMO (which we can add later, as needed), since it is a lot of work and callers of `Commits` will very often not use the value. This means expensive calculations are often wasted.

sqs: SGTM

vcs/repository.go
```diff
@@ -120,6 +120,8 @@ type CommitsOptions struct {
 
 	N    uint // limit the number of returned commits to this many (0 means no limit)
 	Skip uint // skip this many commits at the beginning
+
+	Path string // only commits modifying the given path are selected (optional)
```
sqs: Path can be a file or a dir, right?

vcs/gitcmd/repo.go
```diff
@@ -343,6 +343,10 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
 		args = append(args, "--skip="+strconv.FormatUint(uint64(opt.Skip), 10))
 	}
 
+	if opt.Path != "" {
+		args = append(args, "--follow")
+	}
+
 	// Range
 	rng := string(opt.Head)
```
sqs: Just to be safe, I think it'd be good to check that opt.Head doesn't begin with `-` (in case git options change and this opens an attack vector). Callers might already be doing this, in which case document that callers MUST do this and then no need to do it here.

vcs/gitcmd/repo.go
```diff
@@ -401,6 +409,12 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
 
 	// Count commits.
 	cmd = exec.Command("git", "rev-list", "--count", rng)
+	if opt.Path != "" {
+		// TODO: This doesn't include --follow flag because rev-list doesn't support it, so the number may be slightly off.
+		//       Consider removing the "total" return value from the API altogether, that's why I'm not trying to do more to produce an accurate number here.
+		//       Getting the total number of commits is an expensive operation and likely isn't needed, or deserves a separate dedicated API if needed.
+		cmd = exec.Command("git", "rev-list", "--count", rng, "--", opt.Path)
```
sqs: Just say

```go
cmd.Args=append(cmd.Args, "--", opt.Path)
```

instead of recreating the `exec.Command(...)`

sqs:
Minor comments, feel free to merge when addressed

vcs/repository.go
shurcooL: Yes. Is it clear or should I mention it in comment?

The upstream git docs are more detailed... or so I thought. I tried to look it up again now and I cannot find it saying it can be a file or folder anywhere. It just says this:

> Show only commits that are enough to explain how the files that match the specified paths came to be.

> <paths>
> Commits modifying the given <paths> are selected.

Note that it can be multiple paths (in the git command, but the API change I propose in this PR only allows setting one path max).

The only relevant note I could find is docs for --follow flag:

> --follow
> Continue listing the history of a file beyond renames (works only for a single file).

But I think I verified it works on paths that folders through empirical testing.

vcs/gitcmd/repo.go
shurcooL: Right, I will do one of those two things (either check here, or document that caller is responsible for doing so), thanks!

vcs/gitcmd/repo.go
shurcooL: Ah, nice, you're familiar with `exec.Cmd` internals too. :P I didn't do this because I thought it might be less readable. But it doesn't matter because this code will be deleted soon once we remove total commit count from Commits API.

vcs/repository.go
sqs: In this case I think it's fine to leave as is then

vcs/gitcmd/repo.go
shurcooL: Great, I'll make a followup PR later on!

shurcooL:
Addressed all comments, added tests.

PTAL.

sqs:
LGTM

PR #51, opened 2015-05-21
vcs/gitcmd: Ignore binary files for Search

keegancsmith:
When there is a match in a binary file, the output of `git grep` does not much
what we expect, leading to a slice bounds out of range in
`gitcmd.(*Repository).Search`. The format of a binary match line is

  Binary file HASH:PATH/TO/BINARY/FILE matches


sqs:
LGTM! Thanks!

PR #52, opened 2015-06-02
gitcmd: Don't count all commits unless needed.

shurcooL:
Previously, getting a single commit would always count all commits on that tree, which is expensive and slow. The result was immediately discarded.

I initially tried to add a `calculateTotal bool` parameter, but then I thought I'd try a `total *uint64` as input/output instead. It's probably not the greatest, but it is internal code (not public API) and slightly less redundant. I'm okay with changing it if requested.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -293,7 +293,7 @@ func (r *Repository) getCommit(id vcs.CommitID) (*vcs.Commit, error) {
 		return nil, err
 	}
 
-	commits, _, err := r.commitLog(vcs.CommitsOptions{Head: id, N: 1})
+	commits, err := r.commitLog(vcs.CommitsOptions{Head: id, N: 1}, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -323,7 +323,9 @@ func (r *Repository) Commits(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error
 		return nil, 0, err
 	}
 
-	return r.commitLog(opt)
+	var total uint64
+	commits, err := r.commitLog(opt, &total)
+	return commits, uint(total), err
 }
 
 func isBadObjectErr(output, obj string) bool {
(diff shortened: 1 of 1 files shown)


slimsag:
I can't vet the logic behind it (I don't know this repo's code very well yet), but overall it LGTM. To rid yourself of the pointer you'd have to have an `calculateTotal bool` parameter _and_ `total uint64` return value, so I think the pointer for i/o is alright.

shurcooL:
Thanks for taking a look!

`commitLog` is used in only two places, and it's unexported. Since I've changed its signature, I've updated all places where it's used. One continues to calculate and return the total value, the other previously discarded it and now does that more efficiently.

vcs/gitcmd/repo.go
```diff
@@ -411,23 +414,25 @@ func (r *Repository) commitLog(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, err
...
-	cmd.Dir = r.Dir
-	out, err = cmd.CombinedOutput()
-	if err != nil {
-		return nil, 0, fmt.Errorf("exec `git rev-list --count` failed: %s. Output was:\n\n%s", err, out)
-	}
-	out = bytes.TrimSpace(out)
-	total, err := strconv.ParseUint(string(out), 10, 64)
-	if err != nil {
-		return nil, 0, err
+	if total != nil {
```
nicot: Actually I'm a little confused about this. If a user doesn't care about the information they will leave this as null? And then they will set it to zero if they want to find out the number of commits? Maybe it should just be a separate function. That way callers don't get confused about why it's slow, or why they are passing in an int pointer.

shurcooL: You're looking at code inside `commitLog`. It's unexported, and only has two internal "users". One use on line 296 and the other one line 326.

The public API, and behavior of this library remains completely unchanged.

> If a user doesn't care about the information they will leave this as null?

Yes, see the documentation change on line 339.

nicot: Gotcha, I didn't read the documentation change. Looks good to me then.

PR #53, opened 2015-06-10
vcs/*: fix handling of root paths inside VFS backends

### LINKED ISSUE #23: Rooted paths in vfs?
I'm not sure if the following has a good reason to be exactly this way, or if there's room for improvement.

```Go
import "github.com/sourcegraph/go-vcs/vcs"
import _ "github.com/sourcegraph/go-vcs/vcs/gitcmd"
```

```Go
repo, err := vcs.Open("git", "/some/git-repo")
if err != nil {
	return nil, nil, err
}

var commitId vcs.CommitID = ...

fs, err := repo.FileSystem(commitId)
if err != nil {
	return nil, nil, err
}

fs.Open("/main.go") // doesn't exist
fs.Open("main.go") // works

fs.ReadDir("/") // doesn't exist
fs.ReadDir(".") // works
```

The returned vfs seems to support only relative paths and not absolute paths.

Is there a good reason for that? Or is it that way because that's just how it happened.

Would it be better if it accepted both relative and absolute paths?

slimsag:
This change handles root paths (i.e. `/foo`) as if they were relative
(`foo`) such that asking the VFS for a file at the root works properly.

Without this change it is impossible to use the VFS provided with:

`golang.org/x/tools/godoc/vfs/httpfs`

As `net/http.FileSystem` will implicitly convert all URL paths into
absolute paths at the root of the FS.

Additionally this makes the behavior of the VFS backends comply with
the OS VFS implementation provided by the `godoc/vfs` package.

Fixes #23

### DIFF
--- a/vcs/git/repo.go
+++ b/vcs/git/repo.go
@@ -593,6 +593,8 @@ func (fs *gitFSLibGit2) readFileBytes(name string) ([]byte, error) {
 }
 
 func (fs *gitFSLibGit2) Open(name string) (vfs.ReadSeekCloser, error) {
+	name = util.Rel(name)
+
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
@@ -607,7 +609,7 @@ func (fs *gitFSLibGit2) Lstat(path string) (os.FileInfo, error) {
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
-	path = filepath.Clean(path)
+	path = filepath.Clean(util.Rel(path))
 
 	mtime, err := fs.getModTime()
 	if err != nil {
@@ -636,7 +638,7 @@ func (fs *gitFSLibGit2) Stat(path string) (os.FileInfo, error) {
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
-	path = filepath.Clean(path)
+	path = filepath.Clean(util.Rel(path))
 
 	mtime, err := fs.getModTime()
 	if err != nil {
@@ -761,7 +763,7 @@ func (fs *gitFSLibGit2) ReadDir(path string) ([]os.FileInfo, error) {
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
-	path = filepath.Clean(path)
+	path = filepath.Clean(util.Rel(path))
 
 	var subtree *git2go.Tree
 	if path == "." {
--- a/vcs/git/repo.go
+++ b/vcs/git/repo.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/tools/godoc/vfs"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs/gitcmd"
+	"sourcegraph.com/sourcegraph/go-vcs/vcs/internal"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs/util"
 	"sourcegraph.com/sqs/pbtypes"
 )
@@ -593,7 +594,7 @@ func (fs *gitFSLibGit2) readFileBytes(name string) ([]byte, error) {
 }
 
 func (fs *gitFSLibGit2) Open(name string) (vfs.ReadSeekCloser, error) {
-	name = util.Rel(name)
+	name = internal.Rel(name)
 
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
@@ -609,7 +610,7 @@ func (fs *gitFSLibGit2) Lstat(path string) (os.FileInfo, error) {
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
-	path = filepath.Clean(util.Rel(path))
+	path = filepath.Clean(internal.Rel(path))
 
 	mtime, err := fs.getModTime()
 	if err != nil {
@@ -638,7 +639,7 @@ func (fs *gitFSLibGit2) Stat(path string) (os.FileInfo, error) {
 	fs.repoEditLock.RLock()
 	defer fs.repoEditLock.RUnlock()
 
-	path = filepath.Clean(util.Rel(path))
+	path = filepath.Clean(internal.Rel(path))
 
 	mtime, err := fs.getModTime()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


vcs/util/path.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+// Rel strips the leading "/" prefix from the path string, effectively turning
+// an absolute path into one relative to the root directory.
+func Rel(path string) string {
+	if len(path) > 0 && path[0] == '/' {
+		return path[1:]
+	}
+	return path
```
shurcooL: The 4 lines above are equivalent to `return strings.TrimPrefix(path, "/")`, is that correct?

If so, I think you should just use `strings.TrimPrefix` instead for better readability.

Whenever I see some snippet of Go code like the above, I assume it must be slightly different logic and I spend 20-30 seconds figuring out the nuance that makes it not exactly the same as `strings.TrimPrefix`, and if it turns out to be equivalent, I think "why not just use `strings.TrimPrefix`?"

shurcooL:
First of all, this PR is awesome and thanks for taking on #23!

I want to ask, in which scenarios have you tested the change?

vcs/util/path.go
shurcooL: Additionally, are you sure doing the above is the right thing? What if the input is "/", then the output is a blank string. Would that work?

I know in my vfs adapter that I made as a workaround for #23 I had a special case for "/" inputs and returned "." as output. See [here](https://github.com/shurcooL/go/blob/200c7feca7203b642339097c69bdd4d194b549e1/vfs_util/rooted_vfs.go#L38-L46).

slimsag: >  I think you should just use strings.TrimPrefix instead for better readability.

Yep, good point. I've pushed a commit to address this just now.

> Additionally, are you sure doing the above is the right thing? What if the input is "/", then the output is a blank string. Would that work?

In all of my testing a blank string does work, but my testing didn't include Lstat which it looks like [has a special case in treating "." as "root directory"](https://sourcegraph.com/sourcegraph.com/sourcegraph/go-vcs@72633312692c1d28164bda8da8dc3b814a24b0aa/.tree/vcs/gitcmd/repo.go#startline=891&endline=892), so I'll add a special case for "/".

vcs/util/path.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import "strings"
+
+// Rel strips the leading "/" prefix from the path string, effectively turning
+// an absolute path into one relative to the root directory.
```
shurcooL: Minor, the "/" -> "." special case isn't documented here.

Also, I think this package would be a good candidate to place in `vcs/internal/util` instead of `vcs/util`. It's an internal utility that is used by multiple vcs implementations, but it's not really a part of the API of this repository and we'd want to be free to change/remove it without worrying about external users.

slimsag: Fair points. I've documented the special case and moved the `Rel` function into a new `vcs/internal` package.

I think `vcs/internal` is more canonical than `vcs/internal/util` (we probably won't have more than one internal package, given that it only contains one function at this point).

slimsag:
> First of all, this PR is awesome and thanks for taking on #23!

:smile: 

> I want to ask, in which scenarios have you tested the change?

My testing consisted of running `make test` which all tests passed, and with some code I'm developing that makes use of root `/` paths inside calls to `Stat`, `Open`, and `ReadDir`. However I've only tested this change against the git back-end we use by default (I can't find which one that is ATM).

I feel pretty confident it won't break anything as e.g. even calling `Open` with a rooted path without this change will simply fail, and the only paths this should affect are rooted ones (but of course I'd love help testing it in other cases you can think of!).

vcs/git/repo.go
```diff
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/tools/godoc/vfs"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs"
 	"sourcegraph.com/sourcegraph/go-vcs/vcs/gitcmd"
+	"sourcegraph.com/sourcegraph/go-vcs/vcs/internal"
```
shurcooL: I've never seen this done before, but I think it will work as expected. The internal rule is [defined](https://golang.org/s/go14internal) as follows:

> An import of a path containing the element “internal” is disallowed if the importing code is outside the tree rooted at the parent of the “internal” directory.

The import path contains the element "internal", so it should apply.

slimsag: I saw it used in another project somewhere (I can't remember which one exactly) and just thought it looked a lot cleaner :)

shurcooL:
I've tested both `git` and `gitcmd` backends, and I can confirm that this PR fixes #23.

vcs/internal/path.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "strings"
+
+// Rel strips the leading "/" prefix from the path string, effectively turning
+// an absolute path into one relative to the root directory. A path that is just
+// "/" is treated specially, returning just ".".
+func Rel(path string) string {
+	if path == "/" {
+		return "."
+	}
```
shurcooL: I've done some tests for `git` and `gitcmd`, and I found that this special handling of "/" input and returning "." was not necessary. Everything seems to work just fine if this if statement is removed.

@slimsag, if you can confirm that's the case for you as well, then I suggest let's remove this special handling. It will lead to simpler code and not doing something unnecessary.

shurcooL:
I've made one comment that requires your attention at https://github.com/sourcegraph/go-vcs/pull/53/files#r32350541. Otherwise LGTM!

vcs/internal/path.go
shurcooL: I will retract this suggestion. As discussed, the reason why things continue to work without the special case here is because other code happens to call `filepath.Clean` which replaces empty string with ".".

However, empty string is not a valid path for `os.Open`, for example, while "." is valid. So we should keep this special case for callers that do not use `filepath.Clean`

vcs/internal/path.go
```diff
@@ -0,0 +1,13 @@
+package internal
+
+import "strings"
+
+// Rel strips the leading "/" prefix from the path string, effectively turning
```
shurcooL: Can you please mention that this path string is a "forward-slash separated" path string, not a OS filesystem path string. Meaning it's always going to be "/" for root element, and never "\" or "C:" or something else.

slimsag: Done.

slimsag:
I'll go ahead and merge this as-is and send a follow-up PR to address the ci failure (which is unrelated)

PR #54, opened 2015-06-12
vcs/git: update for recent changes to git2go

slimsag:
This updates the usage of git2go to account for recent API changes in the `next` branch.

Fixes the Travis-ci build.

### DIFF
--- a/vcs/git/clone_update.go
+++ b/vcs/git/clone_update.go
@@ -45,10 +45,7 @@ func init() {
 }
 
 func Clone(url, dir string, opt vcs.CloneOpt) (vcs.Repository, error) {
-	clopt := git2go.CloneOptions{
-		FetchOptions: &git2go.FetchOptions{},
-		Bare:         opt.Bare,
-	}
+	clopt := git2go.CloneOptions{Bare: opt.Bare}
 
 	rc, cfs, err := makeRemoteCallbacks(url, opt.RemoteOpts)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


vcs/git/clone_update.go
```diff
@@ -45,7 +45,10 @@ func init() {
 }
 
 func Clone(url, dir string, opt vcs.CloneOpt) (vcs.Repository, error) {
-	clopt := git2go.CloneOptions{Bare: opt.Bare}
+	clopt := git2go.CloneOptions{
+		FetchOptions: &git2go.FetchOptions{},
```
shurcooL: I think this can just be `nil` if we're not setting any fetch options.

slimsag: Originally I thought that as well, but line 61:

```Go
	if rc != nil {
		clopt.RemoteCallbacks = *rc // this line
	}
```

is now a field of the embedded *FetchOptions struct, so it can't be.

shurcooL: Gotcha. Just realized that too. In that case, would it be more clear to replace that line with this?

```Go
clopt.FetchOptions.RemoteCallbacks = *rc
```

shurcooL: Or better yet, leave it out here completely and:

```Go
if rc != nil {
	clopt.FetchOptions = &git2go.FetchOptions{RemoteCallbacks: *rc}
}
```

shurcooL:
Everything else looks okay to me, LGTM.

vcs/git/clone_update.go
shurcooL: Although that wouldn't scale well if more than one `clopt.FetchOptions` field is being optionally set. Perhaps leaving it as is is best.

slimsag: I usually try to keep initialization in a central place, but in this case I agree with you that is much more clear. Done.

Issue #33, opened 2014-12-22
vcs/gitcmd, vcs/hgcmd: More usage of CombinedOutput for parsing output; should use divided output.


dmitshur:
This is a follow up issue to #30. PR #31 has fixed one instance, but more remain.

There's a WIP branch [`fix/dividedoutput-for-parsing-command-output-thorough`](https://github.com/sourcegraph/go-vcs/compare/fix/dividedoutput-for-parsing-command-output-thorough) that attempts to tackle the rest; it might be used as a base, or for reference.

PR #55, opened 2015-06-22
gitcmd, list directory: ignore error when retrieving url for submodul…

neelance:
…e (submodule may not be initialized)

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -1053,12 +1053,12 @@ func (fs *gitFSCmd) lsTree(path string) ([]os.FileInfo, error) {
 			mode = mode | vcs.ModeSubmodule
 			cmd := exec.Command("git", "config", "--get", "submodule."+name+".url")
 			cmd.Dir = fs.dir
-			out, err := cmd.CombinedOutput()
-			if err != nil {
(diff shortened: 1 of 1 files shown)


vcs/gitcmd/repo.go
```diff
@@ -1053,12 +1053,12 @@ func (fs *gitFSCmd) lsTree(path string) ([]os.FileInfo, error) {
 			mode = mode | vcs.ModeSubmodule
 			cmd := exec.Command("git", "config", "--get", "submodule."+name+".url")
 			cmd.Dir = fs.dir
-			out, err := cmd.CombinedOutput()
-			if err != nil {
-				return nil, fmt.Errorf("invalid `git config` output: %q", out)
+			url := "" // url is not available if submodules are not initialized
+			if out, err := cmd.CombinedOutput(); err == nil {
```
shurcooL: I think this should be `cmd.Output` here. You're not parsing the output for errors, only using it when `err == nil`. But `Stderr` may contain extraneous output (info, warnings), which would then become a part of the parsed `url`.

This is related to an overarching issue #33.

neelance: You're right! I did just move the code. Fixed by 726aeb27281822b3e013dedaad2fef4323753368.

PR #58, opened 2015-06-26
Breaking change: Remove total count from Commits method.

shurcooL:
The total commit count did not respect CommitOptions. In many backends, it performed completely separate, potentially expensive logic.

Many users of Commits discard the total count, resulting in wasted potentially expensive calculations. It makes most sense to have it available in a separate endpoint (with separate options) for interested parties. Begin by removing it from Commits method.

This is a breaking change, but I believe it leads to a better, simpler future API. Comments welcome.

### DIFF
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -217,12 +217,12 @@ func main() {
 			log.Fatal(err)
 		}
 
-		commits, total, err := repo.Commits(vcs.CommitsOptions{Head: commitID, N: 250})
+		commits, err := repo.Commits(vcs.CommitsOptions{Head: commitID, N: 250})
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		fmt.Printf("# Commits (%d total):\n", total)
+		fmt.Println("# Commits:")
 		for _, c := range commits {
 			printCommit(c)
 		}
@@ -366,12 +366,12 @@ func main() {
 			log.Fatal(err)
 		}
 
-		commits, total, err := repo.Commits(vcs.CommitsOptions{Head: commitID, N: 10, Path: path})
+		commits, err := repo.Commits(vcs.CommitsOptions{Head: commitID, N: 10, Path: path})
 		if err != nil {
 			log.Fatal(err)
 		}
 
-		fmt.Printf("# History (%d total):\n", total)
+		fmt.Println("# History:")
 		for _, c := range commits {
 			printCommit(c)
 		}
--- a/vcs/bench_test.go
+++ b/vcs/bench_test.go
@@ -307,7 +307,7 @@ type benchRepository interface {
 	ResolveRevision(string) (vcs.CommitID, error)
 	ResolveTag(string) (vcs.CommitID, error)
 	GetCommit(vcs.CommitID) (*vcs.Commit, error)
-	Commits(vcs.CommitsOptions) ([]*vcs.Commit, uint, error)
+	Commits(vcs.CommitsOptions) ([]*vcs.Commit, error)
 	FileSystem(vcs.CommitID) (vfs.FileSystem, error)
 }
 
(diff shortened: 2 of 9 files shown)


shurcooL:
For reviewers, there are 2 logical commits and I recommend reviewing them separately. The first one is straightforward change to simply remove total and all uses, mentions of it.

The second commit is small but has some important, careful logic changes to optimize the code as a result of the new simpler API. This change should be reviewed with more scrutiny.

gbbr:
I agree and I think this is good. The only time we use the total count is when paginating commits.

On the other hand, I'm not sure having a separate endpoint is a good design. How would you feel about extending the `(vcs).CommitsOptions` structure to contain `Count bool` (or similar) which would trigger this functionality if needed. As a result, the `Commits` method would return:
```go
type CommitList struct {
     Commits []*vcs.Commit
     TotalCount int
}
```
TotalCount can be left at 0 if not requested.

I think most of the times when asking for the total count you are also asking for a (partial) list of commits, so there would be no point in iterating over them twice.

What do you think?

Alternatively, we could leave it as it because I presume that 99% of the times we are using this we also need the total count, and not having that could cause problems. What is your reasoning for this change and what are the gains that you see from this?

shurcooL:
> What is your reasoning for this change and what are the gains that you see from this?

Two rationales:

- Many currently known users of Commits discard the total count (or use a simpler version of it, like `total > len(commits)`), resulting in wasted potentially expensive calculations.
- Simplify API (for both consumers and producers).

For pagination, you'll likely want to request N+1 commits and use that to decide if "Next Page" button should be available.

> I think most of the times when asking for the total count you are also asking for a (partial) list of commits

I can think of specific one use case where you may want to do that. If you want to display a web page with a list of commits and also links to Pages 1, 2, ...  N-2, N-1, N (last page). In what other scenario would you want a partial list of commits when asking for total commit count? It doesn't seem like a very general usage.

> 99% of the times we are using this we also need the total count, and not having that could cause problems.

It's currently underutilized, and the motivation for this PR was to take advantage of that and improve performance.

For example, if a branch contains 10000 commits and you ask for the first 10 commits, currently, it will still iterate over 10000 commits instead of only the first 10.

However, the speedup from doing that is modest. In my testing on a machine with a fast SSD, and a repo with 10000 commits, counting all commits took 72~ ms.

### Closing Words

I don't feel very strongly about this PR. But I do know we currently do not use of the full information that the current API provides, and it is an opportunity to optimize it by taking advantage. Your comments suggest otherwise. I think we need to figure that out.

gbbr:
I agree that it is more efficient to not iterate through all the commits, especially if you say that in most of the cases we don't need the count. But at the same time I also feel that the count should be available, if needed, via the same method, if possible. I've made a suggestions as to how we could do this, but I'm not sure if it is viable.

I'm also not sure how expensive the _"potentially expensive"_ operations are. I don't necessarily want to take leadership of this change and it would be nice to get others' input too - these are just my thoughts along the lines of the PR.

shurcooL:
Those are fair thoughts, thanks for feedback!

> But at the same time I also feel that the count should be available, if needed, via the same method, if possible.

I will think about what a good way to expose that information would be.

beyang:
What about adding a NoCount option that avoids getting the total count if set? That way no breaking changes, but we still get the optimization we want.

shurcooL:
I have considered doing exactly that, but decided to go for the API simplification because I anticipated that almost no one would want to know the total commit count (together with a partial list of commits).

I would still like to learn more about why one would want to get the total commit count when getting a (potentially partial) list of commits. @gbbr can you elaborate on that?

When I started this change, I was under the impression that the negative performance impact was larger than I think it is now. Given that this would be a breaking change and require all downstreams to be updated, I'm not longer sure it's worth it, and I'm happy to close this in favor of going back to less invasive alternative API of adding a `NoTotal` option to `CommitsOptions`. That would help with quantifiably testing the performance improvements this change brings.

PR #59, opened 2015-06-27
Add CommitsOptions.NoTotal field to avoid calculating total.

shurcooL:
This is an alternative implementation of the potential performance improvement proposed in #58. This time it's a non-breaking API change, adding a field that lets users not interested in `total` to opt out of calculating it.

Closes #58 because it offers the same functionality addition, done through a different API (non-breaking).

### DIFF
--- a/vcs/git/repo.go
+++ b/vcs/git/repo.go
@@ -190,8 +190,6 @@ func (r *Repository) GetCommit(id vcs.CommitID) (*vcs.Commit, error) {
 	return r.makeCommit(c), nil
 }
 
-var MaxCommits = 250
-
 func (r *Repository) Commits(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error) {
 	r.editLock.RLock()
 	defer r.editLock.RUnlock()
@@ -235,10 +233,18 @@ func (r *Repository) Commits(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error
 			commits = append(commits, r.makeCommit(c))
 		}
 		total++
-		return total < uint(MaxCommits)
+		// If we want total, keep going until the end.
+		if opt.NoTotal == false {
+			return true
+		}
+		// Otherwise return once N has been satisfied.
+		return (opt.N == 0 || uint(len(commits)) < opt.N)
 	})
 	if err != nil {
 		return nil, 0, err
+	}
+	if opt.NoTotal {
+		total = 0
 	}
 
 	return commits, total, nil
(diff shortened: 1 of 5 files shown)


vcs/git/repo.go
```diff
@@ -235,11 +233,19 @@ func (r *Repository) Commits(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error
 			commits = append(commits, r.makeCommit(c))
 		}
 		total++
-		return total < uint(MaxCommits)
+		// If we want total, keep going until the end.
+		if opt.NoTotal == false {
```
beyang: `if !opt.NoTotal {`

vcs/git/repo.go
```diff
@@ -235,11 +233,19 @@ func (r *Repository) Commits(opt vcs.CommitsOptions) ([]*vcs.Commit, uint, error
 			commits = append(commits, r.makeCommit(c))
 		}
 		total++
-		return total < uint(MaxCommits)
```
beyang: I think it's still a good idea to keep this for now, so we don't regress the perf of any clients that do not pass the `NoTotal parameter`.

vcs/git/repo.go
shurcooL: I find the double negation harder to read "not no-total" than the longer form "no total is false", but might be just me. I'll change it to short form.

vcs/repository.go
```diff
@@ -36,7 +36,7 @@ type Repository interface {
 	// ErrCommitNotFound if no such commit exists.
 	GetCommit(CommitID) (*Commit, error)
 
-	// Commits returns all commits matching the options, as well as
+	// Commits returns all commits matching the options, and optionally
```
beyang: This comment is technically incorrect, as the default computes the total. Should read "Optionally, the caller can request the total not to be computed, as this can be expensive."

beyang:
LGTM pending commetns

vcs/git/repo.go
shurcooL: Ok. Do you know what clients use the `git` backend? I'm only aware of ones that use `gitcmd` backend at this time.

Once they're updated to use `NoTotal`, this performance hack can be reverted. Note that it technically results in incorrect behavior, since if you ask for more than all commits, you'll get only 250 commits max, even if there are more (and of course total will also be capped to 250).

vcs/repository.go
shurcooL: Done in e0b847c8203c7222d46bf554f27c16e36c815179.

vcs/git/repo.go
shurcooL: Changed in e0b847c8203c7222d46bf554f27c16e36c815179.

vcs/git/repo.go
shurcooL: As discussed, there are no known `git` backend users (and if there are and care about this, they can start using `NoTotal` option), so let's side code that is simpler and produces correct (per documentation) results.

PR #61, opened 2015-06-29
Update git2go API.

### LINKED ISSUE #60: git backend needs to be updated for upstream API change.
In https://github.com/libgit2/git2go/commit/d2808d16101e113096b3a1c02e3f91122de74f15 and adjacent commits, the API has changed.

`repo.LookupSubmodule` needs to be replaced with `repo.Submodules.Lookup`, etc.

All travis builds will fail until then.

shurcooL:
Resolves #60.

This PR updates the API. However, it doesn't work now because there seems to be an issue in latest git2go on next branch with Clone method, see https://github.com/libgit2/git2go/commit/4b9cbd78fd266767f6bdf55257c4ee2b1611bbe0#commitcomment-11919367. **Edit:** I've made https://github.com/libgit2/git2go/issues/217 to track that issue.

Until that is resolved, this PR will fail.

### DIFF
--- a/vcs/git/clone_update.go
+++ b/vcs/git/clone_update.go
@@ -75,7 +75,7 @@ func Clone(url, dir string, opt vcs.CloneOpt) (vcs.Repository, error) {
 
 func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
 	// TODO(sqs): allow use of a remote other than "origin"
-	rm, err := r.u.LookupRemote("origin")
+	rm, err := r.u.Remotes.Lookup("origin")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


shurcooL:
libgit2/git2go#217 has been resolved, but another issue https://github.com/libgit2/git2go/issues/218 came up that is preventing tests from passing.

There is an open PR https://github.com/libgit2/git2go/pull/219 that resolves it, waiting on further review and not merged yet.

Skip 2 failing tests for now.

shurcooL:
You can see that the 2nd commit fixed the build here:

https://travis-ci.org/sourcegraph/go-vcs/builds/71881587

slimsag:
A concern: will we encounter those same panics in production (not in tests)?

Otherwise the change LGTM.

shurcooL:
Yes, we may. The latest version of the `git` backend is unusable until that issue is resolved. Vendored older versions should be fine, since the bug was introduced in git2go next branch more recently.

However, were not using the `git` backend at all right now.

This PR turns the `git` backend from "does not compile because invalid API" to "compiles, all tests except 2 pass, and all ok after upstream issue resolved".

The main motivation to merge it is to let Travis build pass so failures on actively used backend (like `gitcmd`) are more visible.

slimsag:
OK, thanks @shurcooL that makes a lot of sense.

LGTM & Merge away! :+1:

PR #62, opened 2015-07-03
add ContainsCommit option for listing branches containing a commit to gitcmd

beyang:
This corresponds to running `git branch --contains {commit-id}`

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -197,6 +197,20 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 		}
 	}
 
+	if opt.ContainsCommit != "" {
+		filteredBranchNames, err := r.branchesWithCommit(opt.ContainsCommit)
+		if err != nil {
+			return nil, err
+		}
+		filteredBranches := make([]*vcs.Branch, 0)
+		for _, branch := range branches {
+			if filteredBranchNames[branch.Name] {
+				filteredBranches = append(filteredBranches, branch)
+			}
+		}
+		branches = filteredBranches
+	}
+
 	if opt.IncludeCommit {
 		for i, branch := range branches {
 			c, err := r.getCommit(branch.Head)
@@ -219,6 +233,30 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 	}
 
 	return branches, nil
+}
+
+func (r *Repository) branchesWithCommit(commitID string) (map[string]bool, error) {
+	cmd := exec.Command("git", "branch", "--contains", commitID)
+	cmd.Dir = r.Dir
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	lines := strings.Split(string(out), "\n")
+	branches := make([]string, 0)
+	for _, line := range lines {
+		if len(line) >= 2 {
+			branches = append(branches, line[2:])
+		}
+	}
+
+	branches_ := make(map[string]bool)
+	for _, branch := range branches {
+		branches_[branch] = true
+	}
+
+	return branches_, nil
 }
 
 func (r *Repository) Tags() ([]*vcs.Tag, error) {
--- a/vcs/repository_test.go
+++ b/vcs/repository_test.go
@@ -427,6 +427,47 @@ func TestRepository_Branches(t *testing.T) {
 	}
 }
 
+func TestRepository_Branches_ContainsCommit(t *testing.T) {
+	t.Parallel()
+
+	gitCommands := []string{
+		"GIT_COMMITTER_NAME=a GIT_COMMITTER_EMAIL=a@a.com GIT_COMMITTER_DATE=2006-01-02T15:04:05Z git commit --allow-empty -m base --author='a <a@a.com>' --date 2006-01-02T15:04:05Z",
+		"GIT_COMMITTER_NAME=a GIT_COMMITTER_EMAIL=a@a.com GIT_COMMITTER_DATE=2006-01-02T15:04:05Z git commit --allow-empty -m master --author='a <a@a.com>' --date 2006-01-02T15:04:05Z",
+		"git checkout HEAD^ -b branch2",
+		"GIT_COMMITTER_NAME=a GIT_COMMITTER_EMAIL=a@a.com GIT_COMMITTER_DATE=2006-01-02T15:04:05Z git commit --allow-empty -m branch2 --author='a <a@a.com>' --date 2006-01-02T15:04:05Z",
+	}
+
+	tests := map[string]struct {
+		repo interface {
+			Branches(vcs.BranchesOptions) ([]*vcs.Branch, error)
+		}
+		commitToExpBranches map[string][]*vcs.Branch
+	}{
+		"git cmd": {
+			repo: makeGitRepositoryCmd(t, gitCommands...),
+			commitToExpBranches: map[string][]*vcs.Branch{
+				"920c0e9d7b287b030ac9770fd7ba3ee9dc1760d9": []*vcs.Branch{{Name: "branch2", Head: "920c0e9d7b287b030ac9770fd7ba3ee9dc1760d9"}},
+				"1224d334dfe08f4693968ea618ad63ae86ec16ca": []*vcs.Branch{{Name: "master", Head: "1224d334dfe08f4693968ea618ad63ae86ec16ca"}},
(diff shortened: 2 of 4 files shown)


vcs/vcs.proto
```diff
@@ -53,6 +53,10 @@ message BranchesOptions {
 	// information against the specified base branch. If left blank, then branches will
 	// not include that information and their Counts will be nil.
 	string behind_ahead_branch = 1 [(gogoproto.moretags) = "url:\",omitempty\""];
+
+	// ContainsCommit specifies a commit hash. If non-empty, branches
```
sqs: Elsewhere we call it a "commit ID", so call it that to be consistent.

In "If non-empty, **branches** will return", you are referring to the `Branches` method, right? Capitalize it.

And I would prefer wording like:

> ContainsCommit filters the list of branches to only those that contain a specific commit ID (if set).

It eliminates ambiguity about whether it filters or adds to the results ("will return" is unclear), and it gets to the point quicker.

vcs/repository_test.go
```diff
@@ -427,6 +427,47 @@ func TestRepository_Branches(t *testing.T) {
...
+		"GIT_COMMITTER_NAME=a GIT_COMMITTER_EMAIL=a@a.com GIT_COMMITTER_DATE=2006-01-02T15:04:05Z git commit --allow-empty -m master --author='a <a@a.com>' --date 2006-01-02T15:04:05Z",
+		"git checkout HEAD^ -b branch2",
+		"GIT_COMMITTER_NAME=a GIT_COMMITTER_EMAIL=a@a.com GIT_COMMITTER_DATE=2006-01-02T15:04:05Z git commit --allow-empty -m branch2 --author='a <a@a.com>' --date 2006-01-02T15:04:05Z",
+	}
+
+	tests := map[string]struct {
+		repo interface {
+			Branches(vcs.BranchesOptions) ([]*vcs.Branch, error)
+		}
+		commitToExpBranches map[string][]*vcs.Branch
```
sqs: Elsewhere we call things `wantXyz` not `expXyz`. Use `commitToWantBranches` for consistency.

vcs/repository_test.go
```diff
@@ -427,6 +427,47 @@ func TestRepository_Branches(t *testing.T) {
...
+			commitToExpBranches: map[string][]*vcs.Branch{
+				"920c0e9d7b287b030ac9770fd7ba3ee9dc1760d9": []*vcs.Branch{{Name: "branch2", Head: "920c0e9d7b287b030ac9770fd7ba3ee9dc1760d9"}},
+				"1224d334dfe08f4693968ea618ad63ae86ec16ca": []*vcs.Branch{{Name: "master", Head: "1224d334dfe08f4693968ea618ad63ae86ec16ca"}},
+				"2816a72df28f699722156e545d038a5203b959de": []*vcs.Branch{{Name: "master", Head: "1224d334dfe08f4693968ea618ad63ae86ec16ca"}, {Name: "branch2", Head: "920c0e9d7b287b030ac9770fd7ba3ee9dc1760d9"}},
+			},
+		},
+	}
+
+	for label, test := range tests {
+		for commit, expBranches := range test.commitToExpBranches {
```
sqs: Elsewhere we use `wantXyz` not `expXyz`. Use `wantBranches` for consistency.

vcs/gitcmd/repo.go
```diff
@@ -221,6 +235,30 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 	return branches, nil
 }
 
+func (r *Repository) branchesWithCommit(commitID string) (map[string]bool, error) {
+	cmd := exec.Command("git", "branch", "--contains", commitID)
```
sqs: For security/paranoia, either use `checkArgSpecSafety` on `commitID`, or use the following line instead:

```go
cmd := exec.Command("git", "branch", "--contains="+commitID)
```

The latter is better IMO.

vcs/gitcmd/repo.go
```diff
@@ -197,6 +197,20 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 		}
 	}
 
+	if opt.ContainsCommit != "" {
+		filteredBranchNames, err := r.branchesWithCommit(opt.ContainsCommit)
+		if err != nil {
+			return nil, err
+		}
+		filteredBranches := make([]*vcs.Branch, 0)
```
sqs: Why call make with a len of 0? Why not just `var filteredBranches []*vcs.Branch`?

vcs/gitcmd/repo.go
```diff
@@ -197,6 +197,20 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
...
+		for _, branch := range branches {
+			if filteredBranchNames[branch.Name] {
```
sqs: I personally prefer making presence checks explicit:

```go
if _, present := filteredBranchNames[branch.Name]; present {
```

That way, you don't have to think, "Can the map value ever be false?"

Then you make the map (and the return type of `branchesWithCommit`) a `map[string]struct{}`, which makes it clear that the return value is a set, not a map.

There are memory savings too, to using a map value type of `struct{}`, I believe, but those matter less.

vcs/gitcmd/repo.go
```diff
@@ -221,6 +235,30 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 	return branches, nil
 }
 
+func (r *Repository) branchesWithCommit(commitID string) (map[string]bool, error) {
+	cmd := exec.Command("git", "branch", "--contains", commitID)
+	cmd.Dir = r.Dir
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, err
```
sqs: I find it very helpful to have descriptive errors, but I see that some other recent contributions don't use them. I still think it's worth it.

```
return nil, fmt.Errorf("exec %v in %s failed: %s (output follows)\n\n%s", cmd.Args, cmd.Dir, err, out)
```

Note that as @shurcooL has noted, this doesn't include stderr. For that reason, you can punt on this and someone can fix error handling all around later.

vcs/gitcmd/repo.go
```diff
@@ -221,6 +235,30 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
...
+	}
+
+	lines := strings.Split(string(out), "\n")
```
sqs: Given that this operation might be called a lot, I think it's worth it to avoid the string conversion.

```go
lines := bytes.Split(out, []byte("\n"))
```

Also, can you preallocate the len of `branches`? When is `len(line) < 2` (if it just occurs once, then you can; otherwise fine to leave as-is).

vcs/gitcmd/repo.go
```diff
@@ -221,6 +235,30 @@ func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
...
+	branches := make([]string, 0)
+	for _, line := range lines {
+		if len(line) >= 2 {
+			branches = append(branches, line[2:])
+		}
+	}
+
+	branches_ := make(map[string]bool)
```
sqs: As suggested above, make this a set (value type `struct{}` not `bool`).

Also, `map[string]bool{}` (or the suggested `map[string]struct{}{}`) is shorter than `make`ing it. But here you know the len, so you should call `make(map[string]struct{}, len(branches))`.

Actually - can you omit the construction of the intermediate slice and just create the map directly in the loop above?

sqs:
Some fairly minor comments.

One big thing, though: it's fine to not add the libgit2 impl, but at least make the libgit2-based `Branches` method return an error if `ContainsCommit` is set. That will prevent incorrect behavior.

shurcooL:
> One big thing, though: it's fine to not add the libgit2/hg impls, but at least make the libgit2/hg-based `Branches` methods return an error if `ContainsCommit` is set. That will prevent incorrect behavior.

The [implementation differences table](https://github.com/sourcegraph/go-vcs#implementation-differences) in the README should also be updated.

vcs/gitcmd/repo.go
shurcooL: It looks like both the intermediate map and slice can be eliminated, unless you really want to do branch name de-duplication. I don't think that should be necessary.

The entire `refs, err := r.showRef("--heads")` step appears to be unnecessary when you have non-empty `ContainsCommit`. If I'm not mistaken, you can just use the result of `git branch --contains <commit_id>` directly as the source of `[]*vcs.Branch`.

PR #63, opened 2015-07-21
gitcmd: Avoid lsTree doing a double lock.

shurcooL:
There are only 2 callers of `lsTree` in this package. Both of them already acquire and hold the `fs.repoEditLock.RLock()` when calling `lsTree`.

Remove `lsTree` doing a double lock and document it as requiring caller to hold the lock.


slimsag:
LGTM

@shurcooL Travis build is failing though -- unrelated -- but just wanted to make sure you were made aware of it in case you weren't already.

shurcooL:
> @shurcooL Travis build is failing though -- unrelated -- but just wanted to make sure you were made aware of it in case you weren't already.

I am aware, thanks. I have described why it fails in #61, merging that PR will fix it.

PR #65, opened 2015-07-21
git: Update git2go API (another instance).

shurcooL:
Followup to #61. This was added to master after, and unfortunately I missed it. This fixes Travis build.


slimsag:
LGTM

PR #68, opened 2015-08-24
Added flag to allow requesting branches merged into a branch.

gbbr:
* Added a feature to Branches method in gitcmd that allows retrieving
merged branches.  
* Refactored code to be less repetitive and reduced branches iteration to 1.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -162,7 +162,7 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 }
 
 // branchFilter is a filter for branch names.
-// If not empty, only contained branch names are allowed. If empty, all name are allowed.
+// If not empty, only contained branch names are allowed. If empty, all names are allowed.
 // The map should be made so it's not nil.
 type branchFilter map[string]struct{}
 
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -161,105 +161,122 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
 	return commitID, nil
 }
 
-// branchCounts returns the behind/ahead commit counts information for branch, against base branch.
-func (r *Repository) branchCounts(branch, base string) (behind, ahead uint, err error) {
-	if err := checkSpecArgSafety(branch); err != nil {
-		return 0, 0, err
-	}
-	if err := checkSpecArgSafety(base); err != nil {
-		return 0, 0, err
-	}
+// branchFilter is a filter for branch names.
+// If not empty, only contained branch names are allowed. If empty, all name are allowed.
+// The map should be made so it's not nil.
+type branchFilter map[string]struct{}
 
-	cmd := exec.Command("git", "rev-list", "--count", "--left-right", fmt.Sprintf("refs/heads/%s...refs/heads/%s", base, branch))
-	cmd.Dir = r.Dir
-	out, err := cmd.Output()
-	if err != nil {
-		return 0, 0, err
+// allows will return true if the current filter set-up validates against
+// the passed string. If there are no filters, all strings pass.
+func (f branchFilter) allows(name string) bool {
+	if len(f) == 0 {
+		return true
 	}
-	behindAhead := strings.Split(strings.TrimSuffix(string(out), "\n"), "\t")
-	b, err := strconv.ParseUint(behindAhead[0], 10, 0)
-	if err != nil {
-		return 0, 0, err
+	_, ok := f[name]
+	return ok
+}
+
+// add adds a slice of strings to the filter.
+func (f branchFilter) add(list []string) {
+	for _, l := range list {
+		f[l] = struct{}{}
 	}
-	a, err := strconv.ParseUint(behindAhead[1], 10, 0)
-	if err != nil {
-		return 0, 0, err
-	}
-	return uint(b), uint(a), nil
 }
 
 func (r *Repository) Branches(opt vcs.BranchesOptions) ([]*vcs.Branch, error) {
 	r.editLock.RLock()
 	defer r.editLock.RUnlock()
 
+	f := make(branchFilter)
+	if opt.MergedInto != "" {
+		b, err := r.branches("--merged", opt.MergedInto)
+		if err != nil {
+			return nil, err
+		}
+		f.add(b)
+	}
+	if opt.ContainsCommit != "" {
+		b, err := r.branches("--contains=" + opt.ContainsCommit)
+		if err != nil {
+			return nil, err
+		}
+		f.add(b)
+	}
+
 	refs, err := r.showRef("--heads")
 	if err != nil {
 		return nil, err
 	}
 
-	branches := make([]*vcs.Branch, len(refs))
-	for i, ref := range refs {
-		branches[i] = &vcs.Branch{
-			Name: strings.TrimPrefix(ref[1], "refs/heads/"),
-			Head: vcs.CommitID(ref[0]),
+	var branches []*vcs.Branch
+	for _, ref := range refs {
+		name := strings.TrimPrefix(ref[1], "refs/heads/")
+		id := vcs.CommitID(ref[0])
+		if !f.allows(name) {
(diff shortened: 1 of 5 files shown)


vcs/gitcmd/repo.go
```diff
@@ -161,105 +161,123 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
+		branches = append(branches, branch)
 	}
-
 	return branches, nil
 }
 
-func (r *Repository) branchesWithCommit(commitID string) (map[string]struct{}, error) {
-	cmd := exec.Command("git", "branch", "--contains="+commitID)
+// branches runs the `git branch` command followed by the given arguments and
+// returns the output as a string.
```
shurcooL: This method returns `[]string`, not `string` as the docs imply.

"and returns the output as a string slice"? Or just "returns the branches and error".

vcs/gitcmd/repo.go
```diff
@@ -161,105 +161,123 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
+func (r *Repository) branches(args ...string) ([]string, error) {
+	cmd := exec.Command("git", append([]string{"branch"}, args...)...)
 	cmd.Dir = r.Dir
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("exec %v in %s failed: %v (output follows)\n\n%s", cmd.Args, cmd.Dir, err, out)
 	}
-
-	lines := bytes.Split(out, []byte("\n"))
+	lines := strings.Split(string(out), "\n")
```
shurcooL: I would consider staying with `lines := bytes.Split(out, []byte("\n"))` here, and converting the lines from `[]byte` to `string` inside the for loop below:

```Go
for i, line := range lines {
	branches[i] = string(line[2:])
}
```

gbbr: My logic was that it might be better to do it once, rather than many times, in a `for` loop. We also don't need the `[]byte` version anywhere in the code, so I saw no use to keep it. If you feel strongly about this or if I've made a mistake in understanding, I am happy to amend.

vcs/gitcmd/repo.go
gbbr: It should say "_returns the branches_". This is a mistake on my end, I forgot to change the documentation (this function did something slightly different originally).

vcs/gitcmd/repo.go
```diff
@@ -161,105 +161,123 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
-// branchCounts returns the behind/ahead commit counts information for branch, against base branch.
-func (r *Repository) branchCounts(branch, base string) (behind, ahead uint, err error) {
-	if err := checkSpecArgSafety(branch); err != nil {
-		return 0, 0, err
-	}
-	if err := checkSpecArgSafety(base); err != nil {
-		return 0, 0, err
-	}
+// stringFilter helps manage a list of string filters.
+type stringFilter struct{ only map[string]struct{} }
```
shurcooL: I like the idea of having this `stringFilter` and using it to filter out branches.

I have a suggestion/idea for how to make its implementation slightly shorter and simpler. It assumes that `stringFilter` is "done" and won't need to change/gain extra features in the future, which seems reasonable given that it's a specialized helper for a specific task here.

If so, consider changing its definition from:

```Go
// stringFilter helps manage a list of string filters.
type stringFilter struct{ only map[string]struct{} }
```

To:

```Go
// branchFilter is a filter for branch names.
// If not empty, only contained branch names are allowed. If empty, all name are allowed.
// The map should be made so it's not nil.
type branchFilter map[string]struct{}
```

Then it's more natural to define it as:

```Go
var f = make(branchFilter)
```

The implementation of `allows` method can remain as is.

But the implementation of `add` method can be changed to:

```Go
func (only branchFilter) add(list []string) {
	for _, l := range list {
		only[l] = struct{}{}
	}
}
```

Basically, removing the `if f.only == nil { f.only = make(map[string]struct{}) }` initialization. That just seems out of place for `add` in my opinion. It feels more explicit to make it once when declaring the filter.

However, that is largely a preference and the difference is quite small, so feel free to go with what you think is better overall.

vcs/gitcmd/repo.go
gbbr: The `string` one actually seems to be better by looking at the implementation, unless I'm missing something.

[`strings.Split`] (https://sourcegraph.com/github.com/golang/go@master/.GoPackage/strings/.def/genSplit) vs [`bytes.Split`](https://sourcegraph.com/github.com/golang/go@master/.GoPackage/bytes/.def/genSplit)

vcs/gitcmd/repo.go
gbbr: I like it! I will definitely use your version.

Fun fact: this was actually called `branchFilter` initially and have recently renamed it to `stringFilter`, though it'd be easier to pick-up :)

vcs/gitcmd/repo.go
shurcooL: That sounds great.

vcs/gitcmd/repo.go
shurcooL: > My logic was that it might be better to do it once, rather than many times, in a `for` loop.

It's a matter of doing it once for the _entire_ string, vs. doing it many times for each individual branch name.

The total sum is about the same, so there's no big difference doing it either way.

I don't feel strongly about this, both ways are okay.

vcs/gitcmd/repo.go
shurcooL: Great. :)

I like `branchFilter` more since it limits the scope to one specific task. There's no need for it to be a `stringFilter` since it's not used anywhere else for filtering strings, so might as well take advantage of that and increase specialization. More general code is harder to work with, so it's best to avoid unless you benefit from it being general - which isn't the case here.

gbbr: :+1: agreed

gbbr:
Addressed all comments, squashed into one commit.

vcs/gitcmd/repo.go
```diff
@@ -161,105 +161,122 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
-// branchCounts returns the behind/ahead commit counts information for branch, against base branch.
-func (r *Repository) branchCounts(branch, base string) (behind, ahead uint, err error) {
-	if err := checkSpecArgSafety(branch); err != nil {
-		return 0, 0, err
-	}
-	if err := checkSpecArgSafety(base); err != nil {
-		return 0, 0, err
-	}
+// branchFilter is a filter for branch names.
+// If not empty, only contained branch names are allowed. If empty, all name are allowed.
```
shurcooL: Minor typo, should be "all name**s** are allowed".

vcs/gitcmd/repo.go
```diff
@@ -161,105 +161,122 @@ func (r *Repository) ResolveTag(name string) (vcs.CommitID, error) {
...
-	}
-
-	branches := make([]*vcs.Branch, len(refs))
-	for i, ref := range refs {
-		branches[i] = &vcs.Branch{
-			Name: strings.TrimPrefix(ref[1], "refs/heads/"),
-			Head: vcs.CommitID(ref[0]),
+	f := make(branchFilter)
+	if opt.MergedInto != "" {
+		b, err := r.branches("--merged", opt.MergedInto)
```
shurcooL: This could be left for another PR, but it seems a little odd that we're doing `"--merged", "<commit>"` and `"--contains=" + "<commit>"` - two different styles of arguments for `git branch` command.

According to http://git-scm.com/docs/git-branch, it looks like only `--merged <commit>` and `--contains <commit>` are supported. But both forms seem to work. Still, it'd be better to change to one style for consistency (I'm in favor of changing `--contains=` to be without `=` liked `--merged`.

shurcooL:
Minor typo, otherwise LGTM and can be merged!

shurcooL:
I've confirmed with @gbbr this is ready and has LGTM from his side. Fixed the typo issue and rebased, merging as soon as Travis passes.

PR #69, opened 2015-08-27
Remove double if guard.

shurcooL:
The `if opt.SSH != nil {` guard was there twice, which is harmless, but unnecessary. Remove it, which simplifies the code.


shurcooL:
This is best to review with ignore-whitespace diff view:

https://github.com/sourcegraph/go-vcs/pull/69/files?w=1

beyang:
LGTM

Issue #30, opened 2014-12-18
vcs/gitcmd: Potential issues in parsing command output due to use of CombinedOutput.


dmitshur:
Consider the following _example_ that demonstrates the issue (thanks to @ktsimpso for helping me find it). The https://github.com/robfig/cron repo has both a tag and branch named "v1". If you `git clone` it and do `git rev-parse` on `refs/tags/v1` you get expected results. However, if you just specify `v1` then you get a warning message printed to stdout:

``` bash
cron $ git rev-parse v1
warning: refname 'v1' is ambiguous.
483223747f6b728cfe1b327fe05b2fa4dd456b27
cron $ git rev-parse v1 2>/dev/null
483223747f6b728cfe1b327fe05b2fa4dd456b27
```

Because of that, I was running into an issue with `vcsstore` that uses `gitcmd` backend when trying to resolve a revision "v1":

`GET http://localhost/git/https/github.com/robfig/cron/.commits/warning:%20refname%20%27v1%27%20is%20ambiguous.%0Ab024fc5ea0e34bc3f83d9941c8d60b0622bfaca4/tree/: 400`

As you can see, the ref `v1` got resolved to `warning:%20refname%20%27v1%27%20is%20ambiguous.%0Ab024fc5ea0e34bc3f83d9941c8d60b0622bfaca4` instead of just `b024fc5ea0e34bc3f83d9941c8d60b0622bfaca4` (which is ambiguous, but that's fine; if I cared to specify if I wanted the tag or branch then I would get the specific one).
## Resolution

In general, it's a good practice to use `Output` rather than `CombinedOutput` when doing machine parsing of command output, because commands may or may not, depending on the situation and user environment, print various warnings or other non-critical information to stderr, which messes up expected output which is being parsed. The exit code should be used to check for errors.

I've had a similar experience of using `CombinedOutput` in my earlier code, which caused a real world issue that a user reported (see https://github.com/shurcooL/gostatus/issues/11). The fix was to switch to `Output`:

https://github.com/shurcooL/go/commit/2b4adbf6c5b812fe3c847032b58d8c717eae4f09

I can see this affects both `vcs/gitcmd` as well as `vcs/hgcmd`. I will make a PR that fixes this for `gitcmd` only at first, mostly because I can actually verify that it works as expected and fixes a real issue. Taking care of `hgcmd` can be another issue.

PR #70, opened 2015-09-02
Fetch list of committers in a repository

pararthshah:
Added method Repository.Committers() to fetch the committers list in a git repo using `git shortlog -sne {rev}` command. The information returned for each committer consists of: name, email, # of commits in the specified branch/revision.

This method is only implemented for `gitcmd`; a mercurial implementation is pending.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -28,14 +28,10 @@ import (
 	"golang.org/x/tools/godoc/vfs"
 )
 
-const (
-	// LogEntryPattern is the regexp pattern that matches entries in the output of
-	// the `git shortlog -sne` command.
-	LogEntryPattern = `\s*([0-9]+)\s+([A-Za-z]+(?:\s[A-Za-z]+)*)\s+<([A-Za-z@.]+)>\s*`
-)
-
 var (
-	logEntryPattern = regexp.MustCompile("^" + LogEntryPattern + "$")
+	// logEntryPattern is the regexp pattern that matches entries in the output of
+	// the `git shortlog -sne` command.
+	logEntryPattern = regexp.MustCompile(`^\s*([0-9]+)\s+([A-Za-z]+(?:\s[A-Za-z]+)*)\s+<([A-Za-z@.]+)>\s*$`)
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


vcs/gitcmd/repo.go
```diff
@@ -27,6 +28,16 @@ import (
 	"golang.org/x/tools/godoc/vfs"
 )
 
+const (
+	// LogEntryPattern is the regexp pattern that matches entries in the output of
+	// the `git shortlog -sne` command.
+	LogEntryPattern = `\s*([0-9]+)\s+([A-Za-z]+(?:\s[A-Za-z]+)*)\s+<([A-Za-z@.]+)>\s*`
```
shurcooL: Why does this need to be a part of API?

vcs/repository.go
```diff
@@ -130,6 +133,14 @@ type CommitsOptions struct {
 	NoTotal bool // avoid counting the total number of commits
 }
 
+// CommittersOptions specifies limits on the list of committers returned by
+// (Repository).Committers.
+type CommittersOptions struct {
+	N int // limit the number of returned committers, ordered by decreasing number of commits (0 means no limit)
+
+	Rev string // the rev for which committer stats will be fetched ("" means HEAD)
```
shurcooL: Does `HEAD` have meaning in general VCS terms, or is it git-specific?

Have you considered making this parameter mandatory, like `Head` is in `CommitsOptions`? Would that be a better or worse API?

vcs/gitcmd/repo.go
```diff
@@ -890,6 +901,45 @@ func (r *Repository) Search(at vcs.CommitID, opt vcs.SearchOptions) ([]*vcs.Sear
...
+		opt.Rev = "HEAD"
+	}
+
+	cmd := exec.Command("git", "shortlog", "-sne", opt.Rev)
+	cmd.Dir = r.Dir
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("exec `git shortlog -sne` failed: %v. Output was:\n\n%s", err, string(out))
+	}
+	out = bytes.TrimSpace(out)
```
shurcooL: You are using `cmd.CombinedOutput()` (which includes stdout and stderr) to get the value of `out` here, and then try to parse it.

There are two reasons to avoid doing that.

1. Stderr should not be necessary. The output you're looking to parse should be completely contained within `stdout`.

2. It can cause issues. Git (and other commands, in general) use stderr to output optional warning messages and the like. If those are present, it will likely cause parsing to fail or return invalid results. See issue #30 where it actually happened and related issue #33 that's still open.

I highly suggest using `cmd.Output()` to capture the value of stdout only, and parsing that. Stderr should not be needed at all, as it's only useful for reporting an error when it occurs. But I don't think it's important enough to spend extra effort capturing it.

pararthshah:
Addressed comments. I'm keeping CommittersOptions.Rev an optional arg since it is better to have the implementations handle that situation. The API clients can expect a reasonable committers list when they don't specify a revision. 'Reasonable' would mean the committers on the default/currently checked out revision. eg. the gitcmd implementation uses `HEAD`, while hgcmd could use `default`.

shurcooL:
Branch `03-hooks` has been merged into master; this PR can now be re-created with `master` as base branch. Closing since it shouldn't be merged into `03-hooks`.

PR #72, opened 2015-09-03
Add HTTPSConfig to RemoteOpts, ability to provide password.

shurcooL:
Use a git pass helper via `GIT_ASKPASS` env var (https://www.kernel.org/pub/software/scm/git/docs/gitcredentials.html) to provide password to git commands.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -94,7 +94,7 @@ func Clone(url, dir string, opt vcs.CloneOpt) (*Repository, error) {
 
 	if opt.HTTPS != nil {
 		env := environ(os.Environ())
-		env.unset("GIT_TERMINAL_PROMPT")
+		env.Unset("GIT_TERMINAL_PROMPT")
 
 		gitPassHelper, err := makeGitPassHelper(opt.HTTPS.Pass)
 		if err != nil {
@@ -630,7 +630,7 @@ func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
 
 	if opt.HTTPS != nil {
 		env := environ(os.Environ())
-		env.unset("GIT_TERMINAL_PROMPT")
+		env.Unset("GIT_TERMINAL_PROMPT")
 
 		gitPassHelper, err := makeGitPassHelper(opt.HTTPS.Pass)
 		if err != nil {
@@ -1255,8 +1255,8 @@ var InsecureSkipCheckVerifySSH bool
 // environ is a slice of strings representing the environment, in the form "key=value".
 type environ []string
 
-// unset unsets a single environment variable.
-func (e *environ) unset(key string) {
+// Unset a single environment variable.
+func (e *environ) Unset(key string) {
 	for i := range *e {
 		if strings.HasPrefix((*e)[i], key+"=") {
 			(*e)[i] = (*e)[len(*e)-1]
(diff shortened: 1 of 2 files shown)


vcs/gitcmd/repo.go
```diff
@@ -614,6 +628,20 @@ func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
 		cmd.Env = []string{"GIT_SSH=" + gitSSHWrapper}
 	}
 
+	if opt.HTTPS != nil {
```
pararthshah: This common logic could be factored out from here and Clone() above.

vcs/gitcmd/repo.go
```diff
@@ -1195,8 +1223,45 @@ func makeGitSSHWrapper(privKey []byte) (sshWrapper, keyFile string, err error) {
...
 // SSH server's certificate or host key. If InsecureSkipCheckVerifySSH
 // is true, the program is susceptible to a man-in-the-middle
 // attack. This should only be used for testing.
 var InsecureSkipCheckVerifySSH bool
+
+// environ is a slice of strings representing the environment, in the form "key=value".
+type environ []string
+
+// unset unsets a single environment variable.
+func (e *environ) unset(key string) {
```
pararthshah: You could use os.Unsetenv to get the same functionality (https://golang.org/pkg/os/#Unsetenv)

pararthshah:
Minor comments, otherwise LGTM.

vcs/gitcmd/repo.go
shurcooL: I can't. I don't want to modify the _current_ environment (of the process running), I want to modify the environment used for for starting a new command. [`os/exec.Cmd.Env`](https://godoc.org/os/exec#Cmd.Env) type is `[]string`.

vcs/gitcmd/repo.go
shurcooL: This would affect both opt.SSH and opt.HTTPS.

It could be, but it's not trivial because it would prevent `defer` from being used as it is used here now.

Additionally, this logic is really important and sensitive. I would prefer to avoid abstracting it away (making it harder to verify correctness), it should be more explicit what happens here. It's only repeated in two places (not 3 or more) at this time, so benefits from abstracting it are lesser.

pararthshah: Fair enough, although IMO makeGitPassHelper could handle the env setup, taking cmd as an arg and setting `cmd.Env = env` internally so that it is called like this:

```
gitPassHelper, env, err := makeGitPassHelper(opt.HTTPS.Pass, cmd)
if err != nil { ... }
defer os.Remove(gitPassHelper)
```

(That would call for renaming the function `makeGitPassHelper` too.)

I feel it's even more important to not duplicate sensitive logic since then there are fewer places it can be screwed up. But I think you could go either way here, so it's totally up to you.

vcs/gitcmd/repo.go
pararthshah: Ah I see. Ignore the comment.

vcs/gitcmd/repo.go
shurcooL: All right, I've looked into it, and I'm not finding a concrete alternative way that is an overall improvement.

Since this PR adds functionality in a consistent way with the existing code, I think it's okay to keep it as is, but we can follow it up with a refactor afterwards.

PR #73, opened 2015-09-18
added FileLister interface and ListFiles method to gitcmd

pararthshah:


shurcooL:
I'll leave some feedback here soon.

vcs/gitcmd/repo.go
```diff
@@ -962,6 +962,31 @@ func (r *Repository) Committers(opt vcs.CommittersOptions) ([]*vcs.Committer, er
...
 
+func (r *Repository) ListFiles(at vcs.CommitID) ([]string, error) {
+	r.editLock.RLock()
+	defer r.editLock.RUnlock()
+
+	if at == "" {
+		at = "HEAD"
+	}
+
+	cmd := exec.Command("git", "ls-files", "--with-tree", string(at))
```
shurcooL: Before passing it to a git command, you should sanitize/verify `at` by using `checkSpecArgSafety`, as is already the case in other places here.

vcs/gitcmd/repo.go
```diff
@@ -962,6 +962,35 @@ func (r *Repository) Committers(opt vcs.CommittersOptions) ([]*vcs.Committer, er
...
+	}
+	out = bytes.TrimSpace(out)
+
+	allFiles := bytes.Split(out, []byte{'\n'})
+	numFiles := len(allFiles)
+	fileNames := make([]string, numFiles)
+	for i := 0; i < numFiles; i++ {
+		fileNames[i] = string(allFiles[i])
+	}
+	return fileNames, nil
```
shurcooL: What do you think about replacing the section above with:

```Go
return strings.Split(string(out), "\n"), nil
```

vcs/gitcmd/repo.go
```diff
@@ -962,6 +962,35 @@ func (r *Repository) Committers(opt vcs.CommittersOptions) ([]*vcs.Committer, er
...
+		at = "HEAD"
+	}
+
+	cmd := exec.Command("git", "ls-files", "--with-tree", string(at))
+	cmd.Dir = r.Dir
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("exec `git ls-files --with-tree %v` failed: %v", at, err)
+	}
+	out = bytes.TrimSpace(out)
```
shurcooL: This feels a little broad and imprecise of a tool for the job. You're just using it to prevent the last element from being a blank line, since the output is newline terminated and that happens when you use `Split(out, "\n")`, right?

How about just doing `out = bytes.TrimSuffix(s, []byte("\n"))` instead?

vcs/gitcmd/repo.go
shurcooL: Have you considered using https://www.kernel.org/pub/software/scm/git/docs/git-ls-tree.html instead? I think it's more direct and returns results you want.

vcs/repository.go
```diff
@@ -175,3 +175,11 @@ type Tags []*Tag
 func (p Tags) Len() int           { return len(p) }
 func (p Tags) Less(i, j int) bool { return p[i].Name < p[j].Name }
 func (p Tags) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+// A FileLister is a repository that can perform actions related to
+// listing the entire file tree.
+type FileLister interface {
+	// ListFiles returns list of all file names in the repo at the
+	// given commit.
```
shurcooL: My biggest concern is with the documentation here. It's extremely unhelpful in that it does not the answer the main question I wanted to see answered when I came here to read it.

Without actually running the command and looking at the output, after reading this, I have no hint as to whether the output is going to include elements like:

- `filename_only.txt` (just the name of file, not full path)
- `full/filepath.txt`
- `/absolute/full/filepath.txt` (rooted path, first character is `/`)
- `filepath\that\is\backslash\separated\on\non\linux\os.txt` (backslash separated path on non-unix OS)

Since go-vcs is meant to be a general abstraction on top of specific VCS systems, and also cross-platform, I think it's important to clarify whether the paths are filepath separator separated (meaning you will get `/` on Linux, OS X, others but `\` on Windows, etc.) or always forward-slash separated paths (so `/` on all OSes regardless of their filepath separator character).

I think it's better to always return consistent results, and go with returning forward-slash separated paths in all cases.

I've done a quick test, and it seems that `git` returns forward-slash separated paths even on Windows, so there's no need to use `filepath.FromSlash`.

So, my suggested changes to the docs would be:

1. Mention that paths are forward-slash-separated.
2. Mention that full paths (including directories, but not rooted) are returned.

A quick example would be very helpful:

> E.g., returned paths have the form "full/path/filename.txt".

shurcooL:
Done reviewing.

vcs/gitcmd/repo.go
shurcooL: Also, have you considered using `-z` flag, and splitting on `byte(0)` character?

vcs/gitcmd/repo.go
shurcooL: Even so, this does not handle the edge case of 0 vs. 1 elements well.

vcs/gitcmd/repo.go
pararthshah: Please explain why is `git ls-tree` more direct than `git ls-files`, since the former command outputs a lot more information which will unnecessarily need to be parsed and discarded before returning the file names (which is what the latter command outputs).

pararthshah: Help me understand why using `-z` flag and splitting on `byte(0)` character is preferable. Is it performance or stability (or both)?

vcs/gitcmd/repo.go
pararthshah: Please explain why `bytes.TrimSpace` is particularly broad and imprecise for this situation, while not being so for 7 other similar usages in this file. Has it got to do with the specific output format of `git ls-files`? I am unable to follow the reasoning here.

vcs/gitcmd/repo.go
pararthshah: Good catch.

vcs/repository.go
pararthshah: Thanks for explaining this lucidly. Totally agree about documenting this correctly. Have incorporated your suggestions.

vcs/gitcmd/repo.go
pararthshah: Not sure what issue you are seeing here, because `strings.Split("", "\n")` returns an empty slice, which is what we'd want to return in case of 0 elements.

vcs/gitcmd/repo.go
shurcooL: My thinking about `git ls-tree` and `git ls-files` is this:

- `git-ls-files` - Show information about files in the index and the working tree
  - This merges the file listing in the directory cache index with the actual working directory list, and shows different combinations of the two.
- `git-ls-tree` - List the contents of a tree object

As I understand it, `git-ls-files` is a superset of `git-ls-tree`. It includes everything tree does, but also augments it with files in current working directory.

go-vcs does not support current working directory, only individual commits.

You'd want to use it with these flags to get the results you want (no extra information):

`git ls-tree --full-tree -r --name-only <commitID>`

shurcooL: > Help me understand why using `-z` flag and splitting on `byte(0)` character is preferable. Is it performance or stability (or both)?

The only difference is in the output:

> Unless the -z option is used, TAB, LF, and backslash characters in pathnames are represented as \t, \n, and \\, respectively.

So it's a matter if we prefer tabs, backslashes or LF characters to be present in the output as is, or converted to their escaped forms.

vcs/gitcmd/repo.go
shurcooL: > because `strings.Split("", "\n")` returns an empty slice

It does not return an empty slice. It returns a slice with 1 element, a blank string.

https://play.golang.org/p/M_wIOtBBg0

shurcooL: > Please explain why `bytes.TrimSpace` is particularly broad and imprecise for this situation, while not being so for 7 other similar usages in this file. Has it got to do with the specific output format of `git ls-files`?

No, it's not specific to `git ls-files`, this comment would apply to other places that use it in the same way.

By "too broad" I mean the `bytes.TrimSpace` operation performs more things than are strictly required. It trims spaces, tabs, newlines, and other characters from the beginning of the input, which is not necessary here. It also trims the same whitespace characters from the end of input, which again is more than strictly required, which is to trim the last single newline character.

Since this is already the style in other places, you don't need to address this as part of this PR, as ideally it can be a refactor that fixes all instances. It's also not critical as behavior remains largely the same for most inputs (unless the first file name begins with a whitespace character, then output is not the same).

gbbr:
Nice job guys! This'll help nicely with fuzzy searching files :)
I think it'd be an awesome idea to remove [this recursivity](https://github.com/sourcegraph/vcsstore/blob/master/vcsclient/fs.go#L200-L224) now that you've added this. It was added solely to get all files. Unfortunately due to an omission on my end the code in `vcsstore` is unusable and slow. I'd recommend removing `Recursive` from `GetFileOptions`.

vcs/gitcmd/repo.go
keegancsmith: I would say it is good practice to always use the null character as a splitter when you have the option. newlines and filesystem paths although rare, are possible.

In the case of ls-tree vs ls-files, @shurcooL makes a good argument why we should use that instead.

pararthshah:
@shurcooL @keegancsmith Thanks for the feedback. I have switched the implementation to use `ls-tree` and split on null character since it is the better way to do it. PTAL.

shurcooL:
LGTM.

PR #76, opened 2015-10-20
Windows support 

alexsaveliev:
- replaced "git rev-parse SPEC^{commit}" with short-hand "git rev-parse SPEC^0" because {} has a special meaning in Windows
- git client expects path arguments to use slashes, so forcing them
- ssh expects path to key file (-i KEY) to use slashes (checked with Cygwin's git and MSYS's git)
- forcing using slashes instead of OS-specific file separator
- making OS-specific GIT_SSH and GIT_ASKPASS helpers (bat files on Windows vs shell scripts on Unix)
- fixed some unit tests (actually everything but "native git" work now on Windows)
- os.Chmod is platform-independent while os.File.Chmod is not implemented on Windows
- On Windows, commit hash may differ when there are symlinks made by different tool (Cygwin/Msys/...) thus computing proper hash before test (affects unit tests)
- code refactoring: extracted method to write a file and set permissions after
- fixed "native hg" stat() method on Windows

TODO list: 
- is there a easy way to make "native git" compile on Windows?
- enhance TestRepository_FileSystem_Symlinks test on Windows - I made some pieces to execute on Unix/Darwin only

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -1269,7 +1269,7 @@ func makeGitSSHWrapper(privKey []byte) (sshWrapper, keyFile string, err error) {
 		return "", keyFile, err
 	}
 
-	tmpFile, err := gitSshWrapper(keyFile, otherOpt)
+	tmpFile, err := gitSSHWrapper(keyFile, otherOpt)
 	return tmpFile, keyFile, err
 }
 
@@ -1315,7 +1315,7 @@ func (e *environ) Unset(key string) {
 }
 
 // Makes system-dependend SSH wrapper
-func gitSshWrapper(keyFile string, otherOpt string) (string, error) {
+func gitSSHWrapper(keyFile string, otherOpt string) (string, error) {
 	// TODO(sqs): encrypt and store the key in the env so that
 	// attackers can't decrypt if they have disk access after our
 	// process dies
@@ -1325,9 +1325,8 @@ func gitSshWrapper(keyFile string, otherOpt string) (string, error) {
 	if runtime.GOOS == "windows" {
 		script = `
 	@echo off
-	ssh "%@"
+	ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "%@"
 `
-		//	ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "%@"
 	} else {
 		script = `
 	#!/bin/sh
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -1240,7 +1240,7 @@ func makeGitSSHWrapper(privKey []byte) (sshWrapper, keyFile string, err error) {
 		return "", "", err
 	}
 	keyFile = kf.Name()
-	err = ioutil.WriteFile(keyFile, privKey, 0600)
+	err = internal.WriteFileWithPermissions(keyFile, privKey, 0600)
 	if err != nil {
 		return "", keyFile, err
 	}
@@ -1266,7 +1266,7 @@ func makeGitPassHelper(pass string) (passHelper string, err error) {
 		script = "#!/bin/sh\necho '" + pass + "'\n"
 	}
 
-	err = ioutil.WriteFile(tmpFile, []byte(script), 0500)
+	err = internal.WriteFileWithPermissions(tmpFile, []byte(script), 0500)
 	return tmpFile, err
 }
 
@@ -1315,6 +1315,6 @@ func gitSshWrapper(keyFile string, otherOpt string) (string, error) {
 		return "", err
 	}
 
-	err = ioutil.WriteFile(sshWrapperName, []byte(script), 0500)
+	err = internal.WriteFileWithPermissions(sshWrapperName, []byte(script), 0500)
 	return sshWrapperName, err
 }
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -1299,16 +1299,17 @@ func gitSshWrapper(keyFile string, otherOpt string) (string, error) {
 	var script string
 
 	if runtime.GOOS == "windows" {
-	script = `
+		script = `
 	@echo off
 	ssh "%@"
-`	
-//	ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "%@"
-} else {
-	script = `
+`
+		//	ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "%@"
+	} else {
+		script = `
 	#!/bin/sh
 	exec /usr/bin/ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "$@"
-`	}
+`
+	}
 
 	sshWrapperName, err := internal.ScriptFile("go-vcs-gitcmd")
 	if err != nil {
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -89,7 +89,7 @@ func Clone(url, dir string, opt vcs.CloneOpt) (*Repository, error) {
 	cmd := exec.Command("git", args...)
 
 	if opt.SSH != nil {
-		gitSSHWrapper, keyFile, err := makeGitSSHWrapper(opt.SSH.PrivateKey)
+		gitSSHWrapper, gitSSHWrapperDir, keyFile, err := makeGitSSHWrapper(opt.SSH.PrivateKey)
 		defer func() {
 			if keyFile != "" {
 				if err := os.Remove(keyFile); err != nil {
@@ -101,6 +101,9 @@ func Clone(url, dir string, opt vcs.CloneOpt) (*Repository, error) {
 			return nil, err
 		}
 		defer os.Remove(gitSSHWrapper)
+		if gitSSHWrapperDir != "" {
+			defer os.RemoveAll(gitSSHWrapperDir)
+		}
 		cmd.Env = []string{"GIT_SSH=" + gitSSHWrapper}
 	}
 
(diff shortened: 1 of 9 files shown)


shurcooL:
> replaced "git rev-parse SPEC^{commit}" with short-hand "git rev-parse SPEC^0" because {} has a special meaning in Windows

I wanted to verify that doing this wouldn't cause a regression for #37, but it appears to be okay.

vcs/internal/script.go
```diff
@@ -0,0 +1,57 @@
...
+func WriteFileWithPermissions(file string, content []byte, perm os.FileMode) error {
+	fi, err := os.OpenFile(file, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, perm)
+	if err != nil {
+		return err
+	}
+	_,err = fi.Write(content)
+	if err != nil {
+		return err
+	}
+	fi.Close()
```
shurcooL: If there's an error during writing, this will return early and the file descriptor will be leaked.

Instead of fixing that, it's better to replace all the code above with just `ioutil.WriteFile(filename, data, perm)`, isn't it? The comment says "Wrapper around ioutil.WriteFile" yet it doesn't use ioutil.WriteFile.

vcs/internal/script.go
```diff
@@ -0,0 +1,57 @@
...
+        	_, err := os.Stat(tempFile)
+        	if err != nil {
+        		if os.IsNotExist(err) {
+        			return filepath.ToSlash(tempFile), nil
+        		} else {
+        			return "", err
+        		}
+        	}
+    	}
+	} else {
```
shurcooL: This does not look `gofmt`ed.

vcs/internal/script.go
alexsaveliev: > ... it's better to replace all the code above with just ioutil.WriteFile(filename, data, perm), isn't it? 

Fixed. However single call of `ioutil.WriteFile` is not enough (it takes int account file permissions only if file didn't exist) so it should be followed by `os.Chmod`. That was the reason of wrapping `ioutil.WriteFile`.

> The comment says "Wrapper around ioutil.WriteFile" yet it doesn't use ioutil.WriteFile.

Yeah, there was `ioutil.WriteFile` but at some point I decided there is something wrong and replaced it with open-write-close-chmod. Restored original idea then

vcs/internal/script.go
alexsaveliev: Formatted code by `gofmt -s`, sorry

alexsaveliev:
>
> I wanted to verify that doing this wouldn't cause a regression for #37
> <https://github.com/sourcegraph/go-vcs/issues/37>, but it appears to be
> okay.
>

https://git-scm.com/docs/git-rev-parse states

...As a special rule, <rev>^0 means the commit itself and is used when
<rev> is the object name of a tag object that refers to a commit object....
so <rev>^0 is identical to <rev>^{commit}.

vcs/internal/script.go
shurcooL: > However single call of ioutil.WriteFile is not enough

Yep, I meant `ioutil.WriteFile` followed by `os.Chmod`, exactly as you've done.

> Yeah, there was `ioutil.WriteFile` but at some point I decided there is something wrong and replaced it with open-write-close-chmod. Restored original idea then

Understandable. Looks good!

vcs/internal/script.go
shurcooL: No problem!

(I do highly recommend configuring your editor to run `gofmt` or better yet `goimports` on-save, if you haven't already. It's very convenient and makes writing Go code significantly more pleasant. All editors have plugins to make this easy.)

shurcooL:
> https://git-scm.com/docs/git-rev-parse states
>
> ...As a special rule, \<rev\>^0 means the commit itself and is used when
> \<rev\> is the object name of a tag object that refers to a commit object....
> so \<rev\>^0 is identical to \<rev\>^{commit}.

Great, it's good that the docs confirm it too.

vcs/gitcmd/repo.go
```diff
@@ -1311,3 +1289,33 @@ func (e *environ) Unset(key string) {
 		}
 	}
 }
+
+// Makes system-dependend SSH wrapper
+func gitSshWrapper(keyFile string, otherOpt string) (string, error) {
```
beyang: s/gitSshWrapper/gitSSHWrapper/

vcs/gitcmd/repo.go
```diff
@@ -1311,3 +1289,33 @@ func (e *environ) Unset(key string) {
...
+	// process dies
+
+	var script string
+
+	if runtime.GOOS == "windows" {
+		script = `
+	@echo off
+	ssh "%@"
+`
+		//	ssh -o ControlMaster=no -o ControlPath=none ` + otherOpt + ` -i ` + filepath.ToSlash(keyFile) + ` "%@"
```
beyang: remove comment

vcs/repository_test.go
```diff
@@ -1152,7 +1175,10 @@ func TestRepository_FileSystem_Symlinks(t *testing.T) {
 			t.Errorf("%s: fs.Lstat(link1): %s", label, err)
 			continue
 		}
-		checkSymlinkFileInfo(label+" (Lstat)", link1Linfo)
+		if runtime.GOOS != "windows" {
+			// TODO(alexsaveliev) make it work
```
beyang: Is this TODO still valid?

vcs/gitcmd/repo.go
alexsaveliev: Done

vcs/gitcmd/repo.go
alexsaveliev: Done

vcs/repository_test.go
alexsaveliev: Yes, this TODO is still valid - at some point we should try fix symlinks functionality on Windows (there are multiple ways with the different outcome of making symlinks: Cygwin's `ln`, MSYS's `ln`, `mklink`) and then make tests run on Windows again

vcs/gitcmd/repo.go
shurcooL: Reference for that: https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

vcs/internal/script.go
```diff
@@ -0,0 +1,53 @@
...
+	gen = rand.NewSource(time.Now().UnixNano())
+}
+
+// Constructs platform-specific temporary script file with a given prefix
+// On Windows such a file must have .bat extension
+func ScriptFile(prefix string) (string, error) {
+
+	if runtime.GOOS == "windows" {
+		for {
+			tempFile := filepath.Join(os.TempDir(), prefix+strconv.FormatInt(gen.Int63(), 36)+".bat")
```
beyang: Can we just use `os.TempFile()` here and forgo the need to generate the random number ourselves?

vcs/repository_test.go
```diff
@@ -1788,3 +1817,27 @@ func appleTime(t string) string {
...
+
+	if git {
+		// git cat-file tree "master^{commit}" | git hash-object -t commit --stdin
+		cat := exec.Command("git", "cat-file", "commit", "master^{commit}")
+		cat.Dir = repoDir
+		hash := exec.Command("git", "hash-object", "-t", "commit", "--stdin")
+		hash.Stdin, _ = cat.StdoutPipe()
+		hash.Stdout = buf
+		hash.Dir = repoDir
+		_ = hash.Start()
```
beyang: you can use `hash.Output()`

vcs/repository_test.go
```diff
@@ -1788,3 +1817,27 @@ func appleTime(t string) string {
...
+		_ = cat.Run()
+		_ = hash.Wait()
+	} else {
+		hash := exec.Command("hg", "--debug", "id", "-i")
+		hash.Dir = repoDir
+		hash.Stdout = buf
+		_ = hash.Run()
```
beyang: you can use `hash.Output()`

vcs/search_test.go
```diff
@@ -36,8 +39,18 @@ func TestRepository_Search_LongLine(t *testing.T) {
 		},
 	}
 
+	// alexsaveliev "echo .... > f1" does not work on Windows, let's create test file different way
+	tmp, err := ioutil.TempFile("", "test")
```
beyang: Can you factor this out into a utility function, called something like `createFileWithContents(contents string)`?

vcs/repository_test.go
beyang: nvm, i see why you did it this way

vcs/repository_test.go
```diff
@@ -1788,3 +1817,27 @@ func appleTime(t string) string {
 	ti, _ := time.Parse(time.RFC3339, t)
 	return ti.Local().Format("200601021504.05")
 }
+
+// Computes hash of last commit in a given repo dir
```
beyang: explain why this is necessary on Windows

vcs/gitcmd/repo.go
```diff
@@ -1335,3 +1313,32 @@ func (e *environ) Unset(key string) {
 		}
 	}
 }
+
+// Makes system-dependend SSH wrapper
```
beyang: s/dependend/dependent/

beyang:
Other than comments, LGTM. Why is Travis failing?

shurcooL:
Travis is failing on Go 1.2 and 1.3, but passing on 1.4 and tip. It's because the `golang.org/x/crypto/ssh` package uses `crypto.Signer` which is available on Go 1.4 and later.

I think we should drop Go 1.2 and 1.3, and use 1.4, 1.5, and tip. 1.3 is more than 1 major version behind current release.

beyang:
yeah, let's go ahead and drop everything below 1.4

vcs/internal/script.go
shurcooL: I think he's doing this because `ioutil.TempFile()` doesn't let you specify the extension, but .bat is needed.

I highly suggest an alternative of using `ioutil.TempDir()` and creating a file with the desired extension inside. See [here](https://github.com/shurcooL/goe/blob/3bcc0e08dda8182005ab103f77e7bf8749d1cf7e/goe.go#L33-L50) for an example.

alexsaveliev: Done as @shurcooL suggested. Thanks for the hint!

vcs/search_test.go
alexsaveliev: Done

vcs/repository_test.go
alexsaveliev: Done

alexsaveliev:
Updated `travis.yml` to include only Go 1.4+

PR #80, opened 2015-11-02
Add UpdateResult to UpdateEverything method.

### LINKED ISSUE #79: Proposal: Add UpdateResult to UpdateEverything method.
This is an early first draft of a proposal. It comes with a sample implementation, but it's very easy to refactor and change it.

#### Background

There are situations where just updating the remote with `UpdateEverything` method and know if it succeeded is not enough, and it's desirable to know a few additional things:

- Has anything changed at all?
- More specifically, have any branches been added, deleted, or updated?

That can enable, e.g., updating the UI when repo updates/changes are available.

# Proposal

I propose having a way to perform a remote update and returning the results of the operation, which contains the changes to branches, if any.

There are 2 ways I see of doing so:

1. Add another interface and method, e.g., `UpdateEverythingResult`, which returns `(UpdateResult, error)`. This is a non-API-breaking change, but it results in two very similar methods.
    - It's possible to then deprecate `UpdateEverything` method and remove it after some time.

2. Change `UpdateEverything` method (and therefore `RemoteUpdater` interface) to have an additional return value, so it returns `(UpdateResult, error)` instead of just `error` as it currently does.

3. (Implicit extra option.) This proposal can be rejected altogether.

I think I would prefer going with the 2nd approach, because this library's API is not yet frozen, and I see this as an overall improvement of the API.

Thoughts?

### Current Implementation

See https://github.com/sourcegraph/go-vcs/pull/80.

shurcooL:
This is an early draft of a sample implementation for proposal #79. It's very easy to refactor and change it.

- Currently implemented in gitcmd backend only, the new functionality parses the output of `git remote update` operation and returns it to the caller.

- This is a breaking API change because it adds an extra return parameter to a method of RemoteUpdater interface.

- Add tests for the parser.

Resolves #79.

### DIFF
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -108,9 +108,11 @@ func main() {
 		log.Printf("Before remote update, HEAD is %s (from %s ago).", preCommit.ID, preCommit.Author.Date)
 
 		log.Printf("Remote-updating repo in dir %s...", dir)
-		if err := repo.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{}); err != nil {
+		result, err := repo.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{})
+		if err != nil {
 			log.Fatal(err)
 		}
+		log.Printf("Result is: %+v\n", result)
 
 		postCommit := getHEADCommit()
 		log.Printf("After remote update, HEAD is %s (from %s ago).", postCommit.ID, postCommit.Author.Date)
--- a/vcs/git/clone_update.go
+++ b/vcs/git/clone_update.go
@@ -73,16 +73,16 @@ func Clone(url, dir string, opt vcs.CloneOpt) (vcs.Repository, error) {
 	return r, nil
 }
 
-func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) (vcs.UpdateResult, error) {
+func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) (*vcs.UpdateResult, error) {
 	// TODO(sqs): allow use of a remote other than "origin"
 	rm, err := r.u.Remotes.Lookup("origin")
 	if err != nil {
-		return vcs.UpdateResult{}, err
+		return nil, err
 	}
 
 	rc, cfs, err := makeRemoteCallbacks(rm.Url(), opt)
 	if err != nil {
-		return vcs.UpdateResult{}, err
+		return nil, err
 	}
 	if cfs != nil {
 		defer cfs.run()
@@ -93,11 +93,11 @@ func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) (vcs.UpdateResult, err
 	}
 
 	if err := rm.Fetch([]string{"+refs/*:refs/*"}, &opts, ""); err != nil {
-		return vcs.UpdateResult{}, err
+		return nil, err
 	}
 
 	// TODO: Calculate value of vcs.UpdateResult.
-	return vcs.UpdateResult{}, nil
+	return nil, nil
 }
 
 type cleanupFuncs []func() error
--- a/vcs/git/clone_update.go
+++ b/vcs/git/clone_update.go
@@ -67,22 +67,22 @@ func Clone(url, dir string, opt vcs.CloneOpt) (vcs.Repository, error) {
 		return nil, err
 	}
 	r := &Repository{Repository: cr, u: u}
-	if err := r.UpdateEverything(opt.RemoteOpts); err != nil {
+	if _, err := r.UpdateEverything(opt.RemoteOpts); err != nil {
 		return nil, err
 	}
 	return r, nil
 }
 
-func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
+func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) (vcs.UpdateResult, error) {
 	// TODO(sqs): allow use of a remote other than "origin"
 	rm, err := r.u.Remotes.Lookup("origin")
 	if err != nil {
-		return err
+		return vcs.UpdateResult{}, err
 	}
 
 	rc, cfs, err := makeRemoteCallbacks(rm.Url(), opt)
 	if err != nil {
-		return err
+		return vcs.UpdateResult{}, err
 	}
 	if cfs != nil {
 		defer cfs.run()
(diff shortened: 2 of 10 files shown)


cmd/go-vcs/go-vcs.go
```diff
@@ -108,9 +108,11 @@ func main() {
 		log.Printf("Before remote update, HEAD is %s (from %s ago).", preCommit.ID, preCommit.Author.Date)
 
 		log.Printf("Remote-updating repo in dir %s...", dir)
-		if err := repo.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{}); err != nil {
+		result, err := repo.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{})
+		if err != nil {
 			log.Fatal(err)
 		}
+		log.Printf("Result is: %+v\n", result)
```
slimsag: Intentional or was this for debugging?

vcs/git/clone_update.go
```diff
@@ -93,10 +93,11 @@ func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
 	}
 
 	if err := rm.Fetch([]string{"+refs/*:refs/*"}, &opts, ""); err != nil {
-		return err
+		return vcs.UpdateResult{}, err
 	}
 
-	return nil
+	// TODO: Calculate value of vcs.UpdateResult.
```
slimsag: needs addressing

vcs/hgcmd/repo.go
```diff
@@ -348,17 +348,18 @@ func (r *Repository) Diff(base, head vcs.CommitID, opt *vcs.DiffOptions) (*vcs.D
...
 	}
 	cmd := exec.Command("hg", "pull")
 	cmd.Dir = r.Dir
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("exec `hg pull` failed: %s. Output was:\n\n%s", err, out)
+		return vcs.UpdateResult{}, fmt.Errorf("exec `hg pull` failed: %s. Output was:\n\n%s", err, out)
 	}
-	return nil
+	// TODO: Calculate value of vcs.UpdateResult.
```
slimsag: needs addressing

cmd/go-vcs/go-vcs.go
shurcooL: Intentional. This command's purpose is for debugging go-vcs.

vcs/git/clone_update.go
shurcooL: How come? I've updated the [README's Implementation Differences](https://github.com/sourcegraph/go-vcs#implementation-differences) table. Do you think that's not enough?

vcs/hgcmd/repo.go
shurcooL: Same reply as https://github.com/sourcegraph/go-vcs/pull/80#discussion_r44209055.

vcs/git/clone_update.go
slimsag: Isn't this for the git implementation, though? Or did I misunderstand this?

vcs/hgcmd/repo.go
slimsag: You're right here, I had missed this was for `hgcmd`. Sorry.

slimsag:
@shurcooL everything you have said makes sense. Sorry for my confusion! This looks good to me. If you feel confident about the change, feel free to merge. If not, feel free to ping another for additional review here.

shurcooL:
> @shurcooL everything you have said makes sense. Sorry for my confusion! This looks good to me.

Great, and no problem!

> If you feel confident about the change, feel free to merge. If not, feel free to ping another for additional review here.

I'm kinda wearing an open-source contributor hat with this PR, and if possible, I'd rather stick to that for the rest of the process. Just learning about what the experience would be like. If it's not possible or takes too long that way, then I'll switch to my project owner hat and do the rest of the work needed to get this merged (and update our dependencies to new API, etc.). :)

vcs/repository_test.go
```diff
@@ -1673,7 +1673,7 @@ func TestRepository_UpdateEverything(t *testing.T) {
 		}
 
 		// update the mirror.
-		err = r.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{})
+		_, err = r.(vcs.RemoteUpdater).UpdateEverything(vcs.RemoteOpts{})
```
shurcooL: It would be better if this test were updated to also check the value of `UpdateResult`. This would be in addition to the unit tests for the update result parser.

shurcooL: Done in a6a26f9.

vcs/remote.go
```diff
@@ -23,5 +23,31 @@ type HTTPSConfig struct {
...
+
+const (
+	// New is a branch that was created.
+	New Operation = iota
+
+	// Updated is a branch that was updated.
+	Updated
+
+	// Deleted is a branch that was deleted.
+	Deleted
```
shurcooL: These names are nice and don't conflict now, but I'm concerned that using up an exported identifier like `New` for an update result operation name might be problematic in the future if a func `New` is desired to be added.

Perhaps adding an `Op` suffix to these would alleviate that potential problem.

shurcooL: Done in 03bd831.

shurcooL:
I've addressed some review comments with the last 2 commits, PTAL.

shurcooL:
Friendly ping.

slimsag:
Sorry for the delays here. I've given this a second more extensive review just now, as it seems I'll be the only reviewer.

Please address adding `TODO` prefixes to the lack-of-implementation comments, after which I'll feel confident in merging this change.

vcs/git/clone_update.go
beyang: What are the semantics of an empty `vcs.UpdateResult`? If it means "nothing changed", then this is technically incorrect, right? If a client happens to be using the `git` implementation with the expectation that if `UpdateResult` is empty, then nothing is changed, then the client might proceed incorrectly. Wouldn't it be better to add a `panic` here, so that if any client tries to call this, it fails? Or alternatively, have an option field in `vcs.RemoteOpts` to request that a result is returned.

shurcooL:
After discussion in #79, there's been a change done in f63f96a to change the interface:

```diff
@@ -23,8 +23,10 @@ type HTTPSConfig struct {
 type RemoteUpdater interface {
 	// UpdateEverything updates all branches, tags, etc., to match the
 	// default remote repository. The implementation is VCS-dependent.
-	// If supported by the implementation, results of update will be returned.
-	UpdateEverything(RemoteOpts) (UpdateResult, error)
+	//
+	// If supported by the implementation, parsed results of update will be returned,
+	// otherwise it'll be nil.
+	UpdateEverything(RemoteOpts) (*UpdateResult, error)
 }
 
 // UpdateResult is the result of parsing output of the remote update operation.
```

Needs a final review and this can be ready to merge.

beyang:
LGTM

shurcooL:
Great, I'll merge this soon when I have a little more bandwidth to update the dependencies.

Unless someone else is blocking on this, then we can merge sooner.

slimsag:
I'll be merging this as part of https://src.sourcegraph.com/sourcegraph/.changes/366

shurcooL:
Woohoo, thanks!

PR #82, opened 2015-11-19
Windows support - fixed pass helper

alexsaveliev:
- removed single quotes around password in pass helper, because Windows echo prints them as well


keegancsmith:
LGTM

keegancsmith:
What happens on windows if the password has two spaces in it? Will echo collapse it to one space?

alexsaveliev:
You are right, PR does not handle situation when password ends with spaces..

alexsaveliev:
Closing in favor of sourcegraph/go-vcs#83

PR #83, opened 2015-11-19
gitcmd: Correctly escape passwords

keegancsmith:
We avoid the need to do unix/batch escaping entirely by just outputting the file
in the script.


keegancsmith:
@alexsaveliev has tested this on windows, and it does not work. Will update as we find a solution.

alexsaveliev:
LGTM

shurcooL:
Thanks for catching and fixing this.

PR #84, opened 2015-12-01
vcs/gitcmd: use 'git remote update --prune' option

slimsag:
Prior to this change we would not use the `--prune` flag which removes
any remote-tracking references that no longer exist on the remote after
updating.

In our app, this caused mirrored repositories updated via `UpdateEverything`
to have branches that cannot ever be deleted (by deleting the branch from
remote).

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -627,7 +627,7 @@ func (r *Repository) UpdateEverything(opt vcs.RemoteOpts) error {
 	r.editLock.Lock()
 	defer r.editLock.Unlock()
 
-	cmd := exec.Command("git", "remote", "update")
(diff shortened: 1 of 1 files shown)


shurcooL:
I thought I had a PR or issue or comment on a commit for this very change, because I noticed `--prune` was missing and thought it should be added. I couldn't find it, so I'm guessing I deferred creating the issue until I could provide reproducible evidence that the flag was needed.

In other words, I'm completely for this change and have wanted to make it myself for a while now, just didn't get around to it.

LGTM.

I've also been using this flag myself locally (e.g., see https://github.com/shurcooL/cmd/commit/928495b164e3a1f8edfcdb3a4f575322e0e91b9d) and it works exactly as expected.

PR #85, opened 2015-12-01
vcs/util/tracer: add a utility package for recording VCS operations to Appdash

slimsag:
This change adds a utility package which records VCS operations to an `*appdash.Recorder`
by wrapping a `vcs.Repository` value. It implements the optional interface types
that a `vcs.Repository` implementation may opt-in to, so no functionality is
lost. Because it works at the interface level we can trace all different
backends (`gitcmd`, `git`, `hg`, etc) without writing new tracing code.

In Appdash, all events are prefixed with `vcs.Type` for easier filtering:

screenshot

### DIFF
--- a/vcs/util/tracer/tracer.go
+++ b/vcs/util/tracer/tracer.go
@@ -1,0 +1,132 @@
+package tracer
+
+import (
+	"fmt"
+	"time"
+
+	"sourcegraph.com/sourcegraph/appdash"
+	"sourcegraph.com/sourcegraph/go-vcs/vcs"
+)
+
+func init() { appdash.RegisterEvent(GoVCS{}) }
+
+// GoVCS records a go-vcs method invocation.
+type GoVCS struct {
+	Name, Args string
+
+	StartTime time.Time
+	EndTime   time.Time
+}
+
+// Schema returns the constant "GoVCS".
+func (GoVCS) Schema() string { return "GoVCS" }
+
+func (e GoVCS) Start() time.Time { return e.StartTime }
+func (e GoVCS) End() time.Time   { return e.EndTime }
+
+// Wrap wraps the given VCS repository, returning a repository which emits
+// tracing events.
+func Wrap(r vcs.Repository, rec *appdash.Recorder) vcs.Repository {
+	// Wrap the repository.
+	t := repository{r: r, rec: rec}
+
+	// Also wrap optional interface.
+	final := vcs.Repository(t)
+	if b, ok := r.(vcs.Blamer); ok {
+		final = blamer{Repository: final, b: b, rec: rec}
+	}
+	if d, ok := r.(vcs.Differ); ok {
+		final = differ{Repository: final, d: d, rec: rec}
+	}
+	if c, ok := r.(vcs.CrossRepoDiffer); ok {
+		final = crossRepoDiffer{Repository: final, c: c, rec: rec}
+	}
+	if f, ok := r.(vcs.FileLister); ok {
+		final = fileLister{Repository: final, f: f, rec: rec}
+	}
+	return final
+}
+
+// blamer implements the vcs.Repository interface, adding a wrapped vcs.Blamer
+// implementation.
+type blamer struct {
+	vcs.Repository
+	b   vcs.Blamer
+	rec *appdash.Recorder
+}
+
+// BlameFile implements the vcs.Blamer interface.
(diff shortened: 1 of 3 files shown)


vcs/util/tracer/tracer.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Wrap wraps the given VCS repository, returning a repository which emits
+// tracing events.
+func Wrap(r vcs.Repository, rec *appdash.Recorder) vcs.Repository {
+	// Wrap the repository.
+	t := repository{r: r, rec: rec}
+
+	// Also wrap optional interface.
+	final := vcs.Repository(t)
+	if b, ok := r.(vcs.Blamer); ok {
```
beyang: Better as a type switch statement:

```
switch r := r.(type) {
  case: vcs.Differ: ...
}
```

vcs/util/tracer/tracer.go
```diff
@@ -0,0 +1,132 @@
...
+	if f, ok := r.(vcs.FileLister); ok {
+		final = fileLister{Repository: final, f: f, rec: rec}
+	}
+	return final
+}
+
+// blamer implements the vcs.Repository interface, adding a wrapped vcs.Blamer
+// implementation.
+type blamer struct {
+	vcs.Repository
```
beyang: Wouldn't this be more robust if the struct embedded the local `repository` struct directly? That way, you'd insure that that all the `vcs.Repository` methods were properly wrapped.

beyang: and likewise for `differ`, `crossRepoDiffer` and `fileLister`?

vcs/util/tracer/vfs.go
```diff
@@ -0,0 +1,104 @@
...
+
+// fileSystem implements the vfs.FileSystem interface.
+type fileSystem struct {
+	fs  vfs.FileSystem
+	rec *appdash.Recorder
+}
+
+// Open implements the vfs.Opener interface.
+func (fs fileSystem) Open(name string) (vfs.ReadSeekCloser, error) {
+	// TODO(slimsag): track time to Close
```
beyang: Why doesn't the `Close()` method track the time currently?

vcs/util/tracer/tracer.go
sourcegraph-test: It's possible I've misunderstood your comment, but I think a type switch cannot work here.

It is the case that `r` could implement all of these interface types (`Blamer`, `Differ`, `CrossRepoDiffer` and `FileLister`). A type switch would match only one of these, whereas the type assertions can match all of them in order.

The goal is to build and return a `vcs.Repository` which wraps all of these optional interfaces only if the input `vcs.Repository` also implements the interfaces.

slimsag: ^ above was me, sorry -- wrong account.

vcs/util/tracer/tracer.go
slimsag: The is partially mentioned in [my other comment](https://github.com/sourcegraph/go-vcs/pull/85#discussion_r46371997), that the goal of these types is to act as an additive wrapper for these optional interfaces. The input `vcs.Repository` might not implement them.

We can't use `repository` directly here because `repository` / `r` is not used for the field; the `final` variable is (which is composed of multiple wrappers, i.e. it could be a `blamer` or a `fileLister` etc, we don't know the concrete type but we do know that they all implement `vcs.Repository`).

vcs/util/tracer/vfs.go
slimsag: Whoops -- it does. Invalid comment.

slimsag:
Dang, sorry for wasting your time here @beyang. This approach won't work for implementing the optional interfaces.

It appears that only the last wrapped one is actually implemented, it doesn't carry through in the embedded e.g. `fileSystem` `vcs.Repository` field like I naively thought it did.

Confirmed with http://play.golang.org/p/PbFdH6v_Ph -- I'm going to have to think up a different approach to wrapping all these optional interfaces.

slimsag:
@beyang I'm not very happy about how it looks, but this implementation will work.

Our only other options would be to pass a `context.Context` around in all of the backends and duplicate the tracing code across every backend. Either way we're looking at a lot of code dup (here or there), and I think this is the best of two evils.

Let me know what you think.

slimsag:
*kind bump*

beyang:
LGTM, sorry about the delay

PR #87, opened 2015-12-21
vcs/gitcmd: unset GIT_ASKPASS prior to appension

slimsag:
or else the new one will not take effect if there is a prior one in play.

This was discovered to be incorrect as part of a [separate / unrelated change](https://src.sourcegraph.com/sourcegraph/.commits/597229995ed0bd8f5e89915cb4cdeb25caa87eb3), but nonetheless was incorrect code.

### DIFF
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -119,6 +119,7 @@ func Clone(url, dir string, opt vcs.CloneOpt) (*Repository, error) {
 		if gitPassHelperDir != "" {
 			defer os.RemoveAll(gitPassHelperDir)
 		}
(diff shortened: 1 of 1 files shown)


keegancsmith:
LGTM

shurcooL:
The same change should be done for the `cmd.Env = []string{"GIT_SSH=" + gitSSHWrapper}` line in the block above.

I'd also suggest just implementing a `Set` method instead of doing `Unset` and then a manual `append`. You can copy a tested implementation from [here](https://gotools.org/github.com/shurcooL/go/osutil?rev=a84e97085eb0de48f81e98e00b1d7e6697f3875d#environ-go-L9-L19).

PR #96, opened 2016-02-04
vcs/git: Make it not a fatal error when git submodule is missing details (URL).

shurcooL:
This allows working with repositories where submodules are not initialized, or the backend doesn't support fetching submodule details.

### DIFF
--- a/vcs/git/vfs.go
+++ b/vcs/git/vfs.go
@@ -153,23 +153,19 @@ func (fs *filesystem) submoduleInfo(path string, e *git.TreeEntry) (*util.FileIn
 	if err != nil {
 		return nil, err
 	}
-	var found *git.Submodule
+	var foundURL string
 	for _, sub := range subs {
 		if sub.Path == path {
-			found = sub
+			foundURL = sub.URL
 			break
 		}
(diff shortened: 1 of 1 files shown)


shurcooL:
This behavior is more in line with how the `gitcmd` backend behaves:

https://github.com/sourcegraph/go-vcs/blob/4c6480c1602e264219517ac7c241c0287ac28640/vcs/gitcmd/repo.go#L1240-L1243

keegancsmith:
LGTM. Is there any chance of a tiny test to confirm both `gitcmd` and `git` have the same behaviour in this case?

shurcooL:
Yeah, I can add a test, but the current tests are failing (and have been for the last week or two), so I'll need to fix that first.

shurcooL:
Talked with @neelance. The entire `vcs/git` backend is in high flux right now and it's not worth spending additional time on it now.

Adding a test here isn't trivial (I'd need to create a git repo in `testdata` that has a non-initialized git submodule).

Given the above, let's punt on that.

keegancsmith:
sure. merge away

PR #97, opened 2016-02-13
cmd/go-vcs: Remove deprecated dependency, improve consistency.

shurcooL:
Remove deprecated "github.com/shurcooL/go/vcs" dependency. That package has been significantly cleaned up and moved to "[github.com/shurcooL/vcsstate](https://github.com/shurcooL/vcsstate)". In the process, the tiny set of functionality (determining the type of a repository in a given directory) was removed since it's available elsewhere ([golang.org/x/tools/go/vcs.FromDir](https://godoc.org/golang.org/x/tools/go/vcs#FromDir)) for most users.

Use a simplified version of golang.org/x/tools/go/vcs.FromDir to determine type of repository.

Expand coverage of dynamic repository type discovery to all methods rather than just a select few (that were added/modified after a certain date).

### DIFF
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -1,4 +1,4 @@
-// The go-vcs program exposes go-vcs's library functionality through a
+// The go-vcs command exposes go-vcs's library functionality through a
 // command-line interface.
 package main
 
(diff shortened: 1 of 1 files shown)


neelance:
LGTM

PR #98, opened 2016-08-20
README: Fix broken links.

shurcooL:
Point to the `vcs.Repository` interface, since that's a better starting point for this project than the root directory (i.e. https://sourcegraph.com/github.com/sourcegraph/go-vcs).

Thanks to @prashantv for reporting the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@ Git and Mercurial (hg).
 
 Note: the public API is experimental and subject to change until further notice.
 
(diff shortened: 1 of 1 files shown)


rothfels:
LGTM

shurcooL:
CI is failing because of some `hg` tests. We need to update Travis config to newer versions and fix the failing tests. I'll follow up on that later.

Issue #11, opened 2014-10-14
Consider using vfs.FileSystem instead of creating identical copy.


dmitshur:
Right now the `go-vcs` package defines its own `FileSystem` interface:

https://sourcegraph.com/github.com/sourcegraph/go-vcs@master/.GoPackage/github.com/sourcegraph/go-vcs/vcs/.def/FileSystem

But it's identical to one already defined in go.tools subrepo:

https://sourcegraph.com/code.google.com/p/go.tools@default/.GoPackage/code.google.com/p/go.tools/godoc/vfs/.def/FileSystem

Despite being identical in signatures, the two interfaces are not compatible due to a technicality:

```
./main.go:79: cannot use fs (type "github.com/sourcegraph/go-vcs/vcs".FileSystem) as type vfs.FileSystem in argument to foo:
    "github.com/sourcegraph/go-vcs/vcs".FileSystem does not implement vfs.FileSystem (wrong type for Open method)
        have Open(string) ("github.com/sourcegraph/go-vcs/vcs".ReadSeekCloser, error)
        want Open(string) (vfs.ReadSeekCloser, error)
```

As long as the interfaces are really the same*, why not just use the existing one?

That is better for the following reason. go.tools is closer to standard Go library, and its `vfs.FileSystem` has [useful helpers](http://godoc.org/code.google.com/p/go.tools/godoc/vfs#pkg-subdirectories). Next, it would make your code slightly shorter. Additionally, people writing new code that targets a virtual file system interface are more likely to pick `vfs.FileSystem`. Otherwise they have to write two versions, or use an adapter such as this one:

``` Go
type VcsToVfs struct {
    vcs.FileSystem
}

func (v *VcsToVfs) Open(s string) (vfs.ReadSeekCloser, error) {
    return v.FileSystem.Open(s)
}
```

``` Go
var fs2 vfs.FileSystem = &VcsToVfs{fs}
```

Which works but wouldn't be necessary if `go-vcs` used `vfs.FileSystem` directly. Consider this a possible enhancement.

(Should the `vfs.FileSystem` interface ever change, as unlikely as that is, you can always _then_ go back to having your own copy if you don't wish to adopt the upstream change.)

PR #101, opened 2018-04-10
all: Add RootType method to vfs.FileSystem implementations.

### LINKED ISSUE #100: vcs/gitcmd does not build on HEAD
```
src/sourcegraph.com/sourcegraph/go-vcs/vcs/gitcmd/repo.go:1026:15: cannot use gitFSCmd literal (type *gitFSCmd) as type vfs.FileSystem in return argument:
	*gitFSCmd does not implement vfs.FileSystem (missing RootType method)
```

shurcooL:
It is needed to continue to implement the [`vfs.FileSystem`](https://godoc.org/golang.org/x/tools/godoc/vfs#FileSystem) interface after golang/tools@16d1af8d88c08b6a22e07a0900aad5279a7fb4fd.

Fixes #100.

### DIFF
--- a/vcs/git/vfs.go
+++ b/vcs/git/vfs.go
@@ -240,6 +240,8 @@ func (fs *filesystem) ReadDir(path string) ([]os.FileInfo, error) {
 	return fis, nil
 }
 
+func (*filesystem) RootType(string) vfs.RootType { return "" }
+
 func (fs *filesystem) String() string {
 	return fmt.Sprintf("git repository %s commit %s (gogit)", fs.dir, fs.oid)
 }
--- a/vcs/gitcmd/repo.go
+++ b/vcs/gitcmd/repo.go
@@ -1267,6 +1267,8 @@ func (fs *gitFSCmd) lsTree(path string) ([]os.FileInfo, error) {
 	return fis, nil
 }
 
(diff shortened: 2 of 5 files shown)


slimsag:
Makes sense. Thanks for sending this @shurcooL !

Let me know when you think this should be merged. I am happy to merge this as-is now, or later.

shurcooL:
Actually, I just updated the PR.

There has been a decision made in https://golang.org/cl/95835 to use empty string as "other" `vfs.RootType`. It will be reflected in upstream documentation soon. But we can use it in this PR already.

So, this is ready to be merged from my side.

keegancsmith:
@shurcooL is the better fix not to just copy the older interface into this repo? The Filesystem interface in the tools repo now seems to be becoming more specific to how go tools work, rather than a generic filsystem interface.

shurcooL:
@keegancsmith I mentioned that in https://github.com/sourcegraph/go-vcs/issues/100#issuecomment-380191873 too. Copying the interface might be a good followup step, but I haven't thought through all the options completely. This PR is just the minimal non-API-breaking change to fix the build.

For reference, this repo used to have a copy of the `vfs.FileSystem` interface back in 2014 but we changed it to use the `godoc` one in issue #11 and PR #12. Using those as reference might be helpful.

PR #103, opened 2018-04-14
Fix CI.

shurcooL:
This PR fixes CI by updating travis config, by dropping Go 1.4, fixing vet and ssh issues, and most importantly, disabling hg implementation tests because they are not functional at this time (see issue #104).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,17 @@
+sudo: false
 language: go
-
 go:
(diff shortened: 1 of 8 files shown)


shurcooL:
This PR is ready now. /cc @sqs @slimsag

keegancsmith:
Thanks!

PR #105, opened 2018-04-27
vcs/gitcmd: Use correct commit range in commitLog.

shurcooL:
[`vcs.CommitsOptions`](https://godoc.org/sourcegraph.com/sourcegraph/go-vcs/vcs#CommitsOptions) has Head and Base fields, documented as:

https://github.com/sourcegraph/go-vcs/blob/d5c79b0b071fb52ee773641c60cfdce93c5f80d0/vcs/repository.go#L128-L129

The Base field is documented to produce results equivalent to `git log Base..Head`. However, that was not the case, the previous code produced results equivalent to `git log Head...Base` instead.

In many cases, those two are equivalent, which is likely why it has gone unnoticed (and the previous tests didn't catch this).

Add more thorough test coverage to catch the subtle difference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: go
 go:
-  - "1.5.x"
   - "1.x"
+  - "1.9.x"
   - master
 matrix:
   allow_failures:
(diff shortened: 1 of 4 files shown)


shurcooL:
It's [failing on Go 1.5](https://travis-ci.org/sourcegraph/go-vcs/builds/371817849) because a dependency uses `math/bits`, which was added in a later version of Go.

shurcooL:
It’s the `golang.org/x/crypto/ssh` package that imports `math/bits`. The `golang.org/x` subrepos only guarantee support for current and previous versions of Go (i.e, 1.10 and 1.9).

So, it seems unavoidable to bump up minimum Go version in Travis to 1.9 as long as the `ssh` package is used. I’ll do that here.

keegancsmith:
Thanks!

PR #109, opened 2018-10-13
Travis: Run tests on Windows.

dmitshur:
There has been work done to the library over time to support Windows.
There's a chance for Windows support to deteriorate over time, since
it's not tested often. Fix that by running tests on Windows in CI.
Travis CI has recently added support for Windows, making this possible.

References:

- https://blog.travis-ci.com/2018-10-11-windows-early-release
- https://twitter.com/travisci/status/1050777251862720513

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 sudo: false
+os:
+  - linux
+  - windows
 language: go
 go:
   - "1.x"


dmitshur:
Looks like it'll take some work to get tests to pass on Windows. I see the following issues so far:

- `gofmt` reports every file as not formatted; this is likely because the [line endings](http://www.cs.toronto.edu/~krueger/csc209h/tut/line-endings.html) of .go files being checked out are converted to from `\n` to `\r\n`. This shouldn't be done.
    - A fix might be to configure the git repo (or Travis) such that the .go files keep their canonical `\n` newlines when checked out on Windows.
    - An alternative fix is to skip `gofmt` on Windows. I suspect this option is worse.
- On Go 1.11.1, `./vcs/...` packages are failing because of `exec: "gcc": executable file not found in %PATH%`.
- On Go 1.9.7, an SSH test in `./vcs` fails because a key file fail to be deleted: `Error removing SSH key file C:\Users\travis\AppData\Local\Temp\go-vcs-gitcmd-key385111935: remove C:\Users\travis\AppData\Local\Temp\go-vcs-gitcmd-key385111935: The process cannot access the file because it is being used by another process.`.

dmitshur:
Looks like @keegancsmith is seeing very similar issues in https://github.com/sourcegraph/go-langserver/pull/329, and has filed an issue about them at https://travis-ci.community/t/go-cant-find-gcc-with-go1-11-1-on-windows/293. This PR can wait until that's resolved.

dmitshur:
This PR is probably obsolete since Travis CI has changed quite a bit over the years and there are newer alternatives available by now. It's a trivial 3 line change that can be re-recreated if needed. Closing.

PR #112, opened 2019-04-06
cmd/go-vcs, vcs/hgcmd: update .../sourcegraph/go-diff/diff import path

dmitshur:
It has changed in https://github.com/sourcegraph/go-diff/pull/30.

Updates https://github.com/sourcegraph/go-diff/issues/27

/cc @sqs

~~You should probably not merge this before https://github.com/sourcegraph/go-diff/issues/33 is resolved, otherwise this PR will break this project in module mode.~~ **Edit:** Issue https://github.com/sourcegraph/go-diff/issues/33 has been resolved, so this is safe to merge now.

### DIFF
--- a/cmd/go-vcs/go-vcs.go
+++ b/cmd/go-vcs/go-vcs.go
@@ -15,7 +15,7 @@ import (
 	"time"
 
 	"github.com/kr/text"
-	"sourcegraph.com/sourcegraph/go-diff/diff"
+	"github.com/sourcegraph/go-diff/diff"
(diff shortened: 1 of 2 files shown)


dmitshur:
Friendly ping @sqs.

This should help anyone who depends on go-vcs, directly or indirectly, similarly to what was mentioned in https://github.com/sourcegraph/go-diff/issues/33#issuecomment-480925293. Please let me know if there any concerns.

Issue #113, opened 2019-08-24
vcs/git: build failure due to import path change in a dependency


dmitshur:
The latest version of the [`sourcegraph.com/sourcegraph/go-vcs/vcs/git`](https://godoc.org/sourcegraph.com/sourcegraph/go-vcs/vcs/git) package fails to build:

```
$ go version
go version go1.12.9 darwin/amd64
$ export GOPATH=$(mktemp -d)
$ go get sourcegraph.com/sourcegraph/go-vcs/vcs/git
/tmp/3p9J1dXg/src/github.com/Unknwon/cae/tz/tz.go:27:2: case-insensitive import collision: "github.com/unknwon/cae" and "github.com/Unknwon/cae"
```

This is because `sourcegraph.com/sourcegraph/go-vcs/vcs/git` imports `sourcegraph.com/sourcegraph/go-git`:

https://github.com/sourcegraph/go-vcs/blob/d784c9520ccdd19883f59efd0a2ae4441f576582/vcs/git/repo.go#L10

And `sourcegraph.com/sourcegraph/go-git` imports `github.com/Unknwon/cae`:

https://github.com/sourcegraph/go-git/blob/5768d0da5d809e85d30fdf69e6fea996d9b35c40/commit_archive.go#L8-L10

But `github.com/Unknwon/cae` has recently changed its import path to `github.com/unknwon/cae` (lower case 'u'):

https://github.com/unknwon/cae/commit/55a0b64484a12115b7322b9eed01e2c26113061d

I'm reporting the problem here rather than in the https://github.com/sourcegraph/go-git issue tracker, because that repository has been archived and is read-only.

The fix is a 3 line change like in https://github.com/smlrepo/go-git/commit/8eddcb03fd7ad072bf7a7ccadd497a327850beee.

/cc @unknwon

keegancsmith:
cc @unknwon

unknwon:
Thanks for the report! Please see https://github.com/sourcegraph/go-vcs/pull/114.

unknwon:
With merge of https://github.com/sourcegraph/go-git/pull/4, this issue should be fixed.

I did a test with command `go get sourcegraph.com/sourcegraph/go-vcs/vcs/git` and don't see the error message anymore.

dmitshur:
It is fixed by https://github.com/sourcegraph/go-git/pull/4, thank you!

PR #114, opened 2019-08-27
Enable Go modules

unknwon:
This PR is meant to fix problem reported in #113.

The easiest fix is to change the import path in `sourcegraph.com/sourcegraph/go-git`, but since it is archived, using Go module `replace` directive is another option.

The [CI will fail for any Go version that does not support Go modules](https://travis-ci.org/sourcegraph/go-vcs/builds/577443962), but pass for "Go: master" (which supports Go modules).


go.mod
```diff
@@ -0,0 +1,27 @@
...
+	gopkg.in/gcfg.v1 v1.2.3 // indirect
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+	sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0
+	sourcegraph.com/sourcegraph/go-git v0.0.0-20160217042419-5768d0da5d80
+	sourcegraph.com/sqs/pbtypes v1.0.0
+)
+
+replace (
+	github.com/Unknwon/cae => github.com/unknwon/cae v0.0.1
+	github.com/Unknwon/com => github.com/unknwon/com v1.0.0
```
dmitshur, line 26: This comment is for information, in case you're not already aware of this.

The replace directives only have effect when this is the main module. They will not help for anyone who has a project that requires this module. See the last paragraph of https://golang.org/wiki/Modules#gomod for more information.

unknwon: Ah... thanks for pointing this out. Seems like the ideal fix is to change import path used in sourcegraph/go-git?

dmitshur: That would be the most direct solution, yes. It would work for both module and GOPATH modes too.

unknwon: Should we unarchive sourcegraph/go-git? What was the reason we archived it and we still use it as a dependency for our own repositories (i.e. this one)? cc @keegancsmith

keegancsmith: We don't actually use go-vcs anymore, so its the sort of thing we would also potentially archive. However, I'll unarchive go-git so you can make the change.

unknwon:
Closed in favor of https://github.com/sourcegraph/go-git/pull/4.

